PR #2, opened 2020-08-18
Add collection utils, with tests

chrisdunelm:
These collection utils are used for convenience throughout the generator.
A imminent future PR will enable CI to run all tests.


chrisdunelm:
I started off using the DS collections, but they are missing quite a bit of
basic functionality, so it became easier to write these. And these allow
much more consistent use than if we'd provided "extensions" to DS.

PR #7, opened 2020-08-24
Initial travis CI

chrisdunelm:


chrisdunelm:
This travis CI run is passing, but the check isn't appearing in this PR. I'll investigate and fix in a future PR.

PR #8, opened 2020-08-25
Code to read custom proto options, plus test

chrisdunelm:
PHP does not support reading custom options directly, so these helper functions are required


src/Utils/ProtoHelpers.php
```diff
@@ -39,4 +44,95 @@ public static function AddProto($desc, $proto)
...
+        static $messageUnknown;
+        if (!$messageUnknown)
+        {
+            $ref = new \ReflectionClass('Google\Protobuf\Internal\Message');
+            $messageUnknown = $ref->getProperty('unknown');
+            $messageUnknown->setAccessible(true);
+        }
+
+        $values = [];
+        if ($message->hasOptions())
```
miraleung, line 61: Nit: Have we considered breaking out parts of this into other methods? Asking because the nesting looks a bit deep.

tests/Utils/CustomOptions.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
```
miraleung, line 1: Nit: Is the UpperCamelCase filename intended? Asking because most protos in googleapis use lower_snake_case.

src/Utils/ProtoHelpers.php
chrisdunelm: Thanks for the thought - I think I'll leave it as-is for now, and will consider splitting it if it becomes too unwieldy later.

tests/Utils/CustomOptions.proto
chrisdunelm: done, thanks.

PR #10, opened 2020-08-25
Very initial AST code; with test

chrisdunelm:


src/Ast/AST.php
```diff
@@ -0,0 +1,90 @@
+<?php declare(strict_types=1);
```
bshaffer, line 1: I'd prefer the `declare(strict_types=1)` to be on its own line and take place after the header. [here](https://github.com/googleapis/google-auth-library-php/blob/v2.x/src/Auth/GoogleAuth.php#L18) is an example.

src/Ast/AST.php
```diff
@@ -0,0 +1,90 @@
...
+
+namespace Google\Generator\Ast;
+
+use Google\Generator\Collections\Vector;
+
+/** Base of the PHP code AST. */
+abstract class AST
```
bshaffer, line 8: Shouldn't casing be consistent between namespace and class name? I am not against it, as i don't mind the difference between class and namespace casing, but I believe it's more common to be consistent.

And the use of `AST` instead of `Ast` is correct according to [this naming RFC](https://wiki.php.net/rfc/class-naming)

src/Ast/AST.php
```diff
@@ -0,0 +1,90 @@
...
+     *     Null values will be ignored.
+     * 
+     * @return AST
+     */
+    public static function block(...$code): AST
+    {
+        $code = Vector::new($code)
+            ->flatten()
+            ->filter(fn($x) => !is_null($x));
+        return new class($code) extends AST
```
bshaffer, line 24: What is the reason for using an anonymous class? Is this because this is more readable than having a different subclass for each block/return/var etc?

src/Ast/AST.php
chrisdunelm: Done.
I think all files currently have it on the first line; I'll do a global change in a future PR.

src/Ast/AST.php
chrisdunelm: I'm not sure it matters too much here - I prefer how reads as all-caps in this specific case, but the namespace looks better to me as Pascal-cased!
I'd probably rather leave this as-is for now; we can always update it later if it's becoming inconsistent with other code in this project.

src/Ast/AST.php
chrisdunelm: Exactly; there's going to be quite a few very simple classes here, and in my opinion it's much more readable to keep them all local in this file.
If some particular anonymous classes become too complex, they can be split out without having to split out all of them into separate files.

src/Ast/AST.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public static function return(Expression $expr): AST
+    {
+        return new class($expr) extends AST
+        {
+            public function __construct($expr)
+            {
+                $this->expr = $expr;
+            }
+            public function ToCode(): string
```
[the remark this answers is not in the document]
chrisdunelm: Oops, my mistake - to many years writing C#! Done.

src/Ast/AST.php
bshaffer: It looks better to me also, I think it works, but wanted to call it out since it's not standard AFAIK

PR #16, opened 2020-08-28
Add test for source-info comments

chrisdunelm:


chrisdunelm:
This test found a bug! The system works :)

PR #23, opened 2020-09-23
Remove Nette/PhpGenerator dependency

chrisdunelm:
Two reasons for this: (1) Nette contained a bug in handling 'use <trait>' statements; (2) Nette forced code to be converted from AST to text at an earlier stage than desirable. Therefore our custom AST has been expanded to cover the 'structural' PHP previously handled by Nette.

### DIFF
--- a/src/Ast/AST.php
+++ b/src/Ast/AST.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 namespace Google\Generator\Ast;
 
 use Google\Generator\Collections\Vector;
+use Google\Generator\Utils\Type;
 
 /** Base of the PHP code AST. */
 abstract class AST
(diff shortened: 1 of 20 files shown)


src/Ast/AST.php
```diff
@@ -73,9 +117,9 @@ public function toCode(): string
 
     /**
      * Create a PHP variable.
-     *
```
jskeet, line 76: It might be worth moving all the whitespace changes into a separate commit, if you have the energy.

jskeet (approved):
I'm no PHP developer, but the structure looks reasonable to me.

src/Ast/AST.php
chrisdunelm: I agree this would be better in a separate commit or PR, but think I'll leave it as-is this time.

amanda-tarafa (approved):
LGTM structure wise. Easy to follow.

PR #24, opened 2020-09-24
Add license header to every source file

chrisdunelm:
And fix trailing whitespace. And remove ClassTypeProxy.php which is no longer required.


chrisdunelm:
This looks big, but it's all completely trivial changes, with no PHP-knowledge required.

jskeet (approved):

PR #26, opened 2020-09-25
Add generation of 'getClientDefaults()' method

chrisdunelm:
(There's also a small amount of code for handing method parameters which isn't used in this PR, but will be used in the next PR)


amanda-tarafa (approved):
LGTM structure wise.

PR #31, opened 2020-09-29
Tests use files to define expected generated code

chrisdunelm:


tests/ProtoTests/ProtoTest.php
```diff
@@ -27,181 +27,27 @@ final class ProtoTest extends TestCase
...
-            ],
-        ];
+        // Conventions:
+        // * The proto package is 'testing.<proto-name>'.
+        // * The expected file contents are based in the same directory as the proto file.
+        $descBytes = $this->loadDescriptorBytes("ProtoTests/{$protoPath}");
+        $package = 'testing.' . basename($protoPath, '.proto');
+        $codeIterator = CodeGenerator::GenerateFromDescriptor($descBytes, $package);
+
+        foreach ($codeIterator as [$relativeFilename, $code]) {
```
jskeet, line 39: One nit: an empty code generator would still pass this test, because we don't check that all the expected files are actually generated.

jskeet (approved):
Much better than inlining - one issue in terms of the symmetry.

tests/ProtoTests/ProtoTest.php
chrisdunelm: Good point, thanks. TODO added.

PR #37, opened 2020-09-30
Include code example in phpdoc

chrisdunelm:


src/Ast/AST.php
```diff
@@ -426,6 +426,36 @@ public function toCode(): string
...
+            {
+                return $this->clone(fn($clone) => $clone->finallyCode = AST::block(...$finallyCode));
+            }
+            public function toCode(): string
+            {
+                return
+                    "try {\n" .
+                    static::toPhp($this->tryCode) . "\n" .
+                    '}' .
+                    (is_null($this->finallyCode) ? '' : "finally {\n" . static::toPhp($this->finallyCode) . "\n}");
```
jskeet, line 454: What does try without catch or finally do, out of interest?

jskeet (approved):

src/Ast/AST.php
chrisdunelm: Nothing, it's invalid ;) So it'll always be required to add a `finally` or `catch` part. But so far I haven't needed to generate any `catch`es, so that code isn't yet implemented.

PR #42, opened 2020-10-01
Initial execution of generated test

chrisdunelm:
These are all the changes required to successfully execute the generated tests, which now execute as part of the default unit-test run.
Apologies that the changes are a bit all over the place.


chrisdunelm:
Please wait to review. I'll find out why this test is failing (naturally it passes on my local machine!)

src/Ast/AST.php
```diff
@@ -48,20 +48,29 @@ protected static function deref($obj): string
         return $obj === static::SELF || $obj instanceof ResolvedType ? '::' : '->';
     }
 
-    protected static function toPhp($x): string
+    protected static function toPhp($x, &$omitSemicolon = false): string
```
jskeet, line 51: It's hard to know without having more context, but is this better than `appendSemiColon = true`? I tend to prefer positive variables instead of negative ones, but there are definitely exceptions to that.

src/Ast/AST.php
```diff
@@ -48,20 +48,29 @@ protected static function deref($obj): string
...
+                return $x;
             } else {
                 // Otherwise strings are treated as string literals.
                 return "'{$x}'";
             }
         } elseif (is_numeric($x)) {
             return strval($x);
         } elseif ($x instanceof PhpClassMember) {
-            return $x->getName();
+            if ($x instanceof PhpProperty) {
```
jskeet, line 69: Use a conditional?

jskeet (approved):

jskeet:
Too late, already reviewed - hadn't noticed the failure ;)

src/Ast/AST.php
chrisdunelm: I tend to prefer a parameter with a default value to use `false` as the default value; but I admit it's not always clear which is the easiest to read. I'll this this as-is for now.

src/Ast/AST.php
chrisdunelm: done

chrisdunelm:
Frustratingly I can't get these errors to reproduce locally; and they look like genuine errors, so I'm confused why they pass on my machine.
I'll continue investigating.

chrisdunelm:
Woohoo!

PR #46, opened 2020-10-06
Refactor method-details to allow different method types

chrisdunelm:
And introduce the very beginnings of LRO

### DIFF
--- a/tests/ProtoTests/ProtoTest.php
+++ b/tests/ProtoTests/ProtoTest.php
@@ -47,8 +47,14 @@ final class ProtoTest extends TestCase
         // TODO: Check that all expected files are actually generated!
     }
 
-    public function testProtoBasic(): void
+    public function testBasic0(): void
     {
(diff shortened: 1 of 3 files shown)


tests/ProtoTests/ProtoTest.php
```diff
@@ -47,8 +47,14 @@ private function runProtoTest(string $protoPath): void
         // TODO: Check that all expected files are actually generated!
     }
 
-    public function testProtoBasic(): void
+    public function testBasic0(): void
```
amanda-tarafa, line 50: nit: there's a 0 at the end of this method name?

amanda-tarafa (approved):
LGTM

tests/ProtoTests/ProtoTest.php
chrisdunelm: Yes :) The `0` is there to make this test name _not_ be a prefix of other test names, meaning it's easy to run just this test!
The same is done in the C# tests.

PR #47, opened 2020-10-12
Generate LRO helper methods in client class

chrisdunelm:
Methods that require generating in the client clas if there are any LRO methods in this client service.

Further codegen and test execution will be in a future PR.


chrisdunelm:
The files in src/Google/LongRunning/... are generated from protoc, so can be mostly ignored.

src/Ast/AST.php
```diff
@@ -461,6 +461,33 @@ public function toCode(): string
         };
     }
 
+    /**
+     * Create a ternary expression.
```
jskeet, line 465: Well, creates a conditional ?: expression.
You know my feelings about using the arity of an expression as its name ;)

jskeet (approved):
Structure-only review, but looks basically fine.

src/Generation/MethodDetails.php
```diff
@@ -39,8 +45,24 @@ private static function createLro(ServiceDetails $svc, MethodDescriptorProto $de
             public function __construct($svc, $desc)
             {
                 parent::__construct($svc, $desc);
-                // TODO: Load LRO-specific details.
+                $this->methodType = MethodDetails::LRO;
+                $this->isLro = true;
+                $catalog = $svc->catalog;
+                $lroData = ProtoHelpers::getCustomOption($desc, CustomOptions::GOOGLE_LONGRUNNING_OPERATIONINFO, OperationInfo::class);
+                if (is_null($lroData)) {
+                    throw new \Exception('An LRO method must provide an `google.api.operation` option.');
```
amanda-tarafa, line 53: nit:
```suggestion
                    throw new \Exception('An LRO method must provide a `google.api.operation` option.');
```

tests/ProtoTests/BasicLro/Gapic/BasicLroGapicClient.php
```diff
@@ -0,0 +1,214 @@
...
+     *
+     * @return OperationResponse
+     *
+     * @experimental
+     */
+    public function resumeOperation($operationName, $methodName = null)
+    {
+        $options = isset($this->descriptors[$methodName]['longRunning']) ? $this->descriptors[$methodName]['longRunning'] : [];
+        $operation = new OperationResponse($operationName, $this->getOperationsClient(), $options);
+        $operation->reload();
```
amanda-tarafa, line 109: Just to double check that reload doesn't return a new OperationResponse.

amanda-tarafa (approved):
LGTM

miraleung (approved):

src/Ast/AST.php
chrisdunelm: Comment changed :)

src/Generation/MethodDetails.php
chrisdunelm: Done, thanks

tests/ProtoTests/BasicLro/Gapic/BasicLroGapicClient.php
chrisdunelm: Thanks for checking - this code is correct in that `reload` does not return a new OperationResponse (unlike C#).
This line is modelled from the monolithic generator here: https://github.com/googleapis/gapic-generator/blob/ae9338ab74e0fd91ee2814b348d59562693a76a6/src/main/resources/com/google/api/codegen/php/client_impl.snip#L303

PR #53, opened 2020-10-14
Improved fake messages for testing

chrisdunelm:


tests/FakeMessage.php
```diff
@@ -0,0 +1,45 @@
...
+use Google\Protobuf\Internal\Message;
+
+/**
+ * Fake message base class for tests, that emulates any proto message well enough for testing.
+ */
+abstract class FakeMessage extends Message
+{
+    public function __construct() { }
+
+    public function __call(string $name ,array $arguments)
```
jskeet, line 30:
```suggestion
    public function __call(string $name, array $arguments)
```

tests/FakeMessage.php
```diff
@@ -0,0 +1,45 @@
...
+        $prefix = substr($name, 0, 3);
+        $suffix = substr($name, 3);
+        switch ($prefix) {
+            case 'set':
+                $this->$suffix = $arguments[0];
+                return;
+            case 'get':
+                return $this->$suffix;
+            default:
+                throw new \Exception("FakeMessage cannot handle method call: '{$name}'");
```
jskeet, line 42: If we're guarding against that, guard against method names shorter than three characters?

jskeet (approved):

tests/FakeMessage.php
chrisdunelm: This is OK, as substr() returns a truncated or empty string if $name is too short. The switch then fails as expected.
Comment added.

PR #57, opened 2020-10-14
Complete LRO test generation for simple RPC

chrisdunelm:


chrisdunelm:
The formatting error in the generated code is a known problem, and will be fixed in a future PR.

src/Generation/UnitTestsGenerator.php
```diff
@@ -309,12 +309,49 @@ private function testExceptionalCaseLro(MethodDetails $method): PhpMethod
...
+        $response = AST::var('response');
+        $expectedOperationsRequestObject = AST::var('expectedOperationsRequestObject');
+        $ex = AST::var('ex');
+        [$initCode, $operationsTransport, $client, $transport] = $this->lroTestInit($method->testExceptionMethodName);
         return AST::method($method->testExceptionMethodName)
             ->withAccess(Access::PUBLIC)
             ->withBody(AST::block(
                 $initCode,
-                // TODO: Complete test...
+                AST::assign($status, AST::new($this->ctx->type(Type::stdClass()))()),
```
dwsupplee, line 323: Would you be open to formatting like the following? At least for me it makes parsing what is going on a little easier, and seems like it may make the file a little more friendly when it needs to be edited.

```php
AST::assign(
    $status,
    AST::new($this->ctx->type(Type::stdClass()))()
),
AST::assign(
    AST::access(
        $status,
        AST::property('code')
    ),
    AST::access(
        $this->ctx->type(Type::fromName(Code::class)),
        AST::constant('DATA_LOSS')
    )
),
```

src/Generation/UnitTestsGenerator.php
```diff
@@ -309,12 +309,49 @@ private function testExceptionalCaseLro(MethodDetails $method): PhpMethod
...
+                    $response->pollUntilComplete(AST::array([
+                        'initialPollDelayMillis' => 1,
+                    ])),
+                    '// If the pollUntilComplete() method call did not throw, fail the test',
+                    ($this->fail)('Expected an ApiException, but no exception was thrown.'),
+                )->catch($this->ctx->type(Type::fromName(ApiException::class)), $ex,
+                    ($this->assertEquals)(AST::access($status, AST::property('code')), $ex->instanceCall(AST::method('getCode'))()),
+                    ($this->assertEquals)($expectedExceptionMessage, $ex->instanceCall(AST::method('getMessage'))()),
+                ),
+                '// Call popReceivedCalls to ensure the stubs are exhausted',
```
dwsupplee, line 350: The line formatting in the generated file seems to be a bit off.

dwsupplee: Oops, I just realized this is what you were referring to in the description - nevermind!

tests/ProtoTests/BasicLro/Tests/BasicLroClientTest.php
```diff
@@ -118,5 +121,36 @@ public function method1ExceptionTest()
...
+            $response->pollUntilComplete([
+                'initialPollDelayMillis' => 1,
+            ]);
+            // If the pollUntilComplete() method call did not throw, fail the test
+            $this->fail('Expected an ApiException, but no exception was thrown.');
+        } catch (ApiException $ex) {
+            $this->assertEquals($status->code, $ex->getCode());
+            $this->assertEquals($expectedExceptionMessage, $ex->getMessage());
+        }
+// Call popReceivedCalls to ensure the stubs are exhausted
```
miraleung, line 150: Nit: Align with lines below?

miraleung (approved):

tests/ProtoTests/BasicLro/Tests/BasicLroClientTest.php
chrisdunelm: Yes, this formatting is wrong ;)
This is generated code, and the problem is due to the code formatter being used, and there's a TODO to fix it. For the moment I'll leave it as it is, but it'll definitely get fixed soonish.

src/Generation/UnitTestsGenerator.php
chrisdunelm: My opinion is that I'd rather not do this, as it will cause the source code to get considerably longer.
I suggest leaving as it is for now, but we can always reformat later if it becomes a problem.

PR #59, opened 2020-10-21
Add license and generated-code headers

chrisdunelm:


src/Main.php
```diff
@@ -31,7 +31,8 @@
 // Generate PHP code.
 // At the moment $files is just the file content.
 // TODO: Change this to be file location and content
-$files = CodeGenerator::GenerateFromDescriptor($descBytes, $package);
+$year = (int)date('Y');
```
jskeet, line 34: Critically: which time zone is this using? (I'm kidding...)

jskeet (approved):

src/Main.php
chrisdunelm: :)

PR #66, opened 2020-11-05
Refactor: Move example code generation into their own file/class

chrisdunelm:


src/Generation/GapicClientGenerator.php
```diff
@@ -46,11 +46,14 @@ public static function generate(SourceFileContext $ctx, ServiceDetails $serviceD
 
     private SourceFileContext $ctx;
     private ServiceDetails $serviceDetails;
+    private GapicClientExamplesGenerator $examples;
+    private bool $hasLro;
```
jskeet, line 50: I don't understand why this is a new field, when you seemed to already be using it, but hey...

jskeet (approved):

src/Generation/GapicClientGenerator.php
chrisdunelm: Yes, this change was not to do with this PR really.
In PHP declaring class fields is optional, but I generally prefer to do it, and this one was missing so I added it in this PR (which perhaps I shouldn't have).

PR #69, opened 2020-11-05
ServerStreaming generation, including test generation

chrisdunelm:


jskeet (approved):
Can't say I follow all of it, but it seems structurally fine.

PR #70, opened 2020-11-05
Refactor test generation - make consistent the order of methods

chrisdunelm:


chrisdunelm:
This is a simple re-ordering of methods, nothing else.

jskeet (approved):

PR #74, opened 2020-11-10
Initial integration test implementation

chrisdunelm:
This compares the monolithic-generator and micro-generator outputs and only allows whitespace and trailing-comma changes.

Only a single generated file is currently compared; future PRs will add all other generated files, and other proto inputs.

### DIFF
--- a/src/Main.php
+++ b/src/Main.php
@@ -24,15 +24,33 @@ error_reporting(E_ALL);
 // TODO: Support running as protoc plugin.
 // TODO: Provide help/usage if incorrect command-line args provided.
 // Read command-line args.
-$opts = getopt('', ['descriptor:', 'package:']);
+// TODO: Include cmd-line arg for grpc-service-config.
+$opts = getopt('', ['descriptor:', 'package:', 'output:']);
+if (!isset($opts['descriptor']) || !isset($opts['package'])) {
+    print("Invalid arguments. Expect:\n");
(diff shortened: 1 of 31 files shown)


chrisdunelm:
Apologies that this is a moderately large PR.
However, all the changes to the test files are purely mechanical.

src/Main.php
```diff
@@ -24,15 +24,33 @@
...
 // Generate PHP code.
 $year = (int)date('Y');
-$files = CodeGenerator::generateFromDescriptor($descBytes, $package, $year);
-foreach ($files as [$relativeFilename, $fileContent]) {
-    // TODO: Later this won't just print out the generated file content.
-    print("File: '{$relativeFilename}':\n");
-    print($fileContent . "\n");
+$files = CodeGenerator::generateFromDescriptor($descBytes, $package, $year, null);
+if (is_null($outputDir)) {
+    // TODO: Remove printout; only save files to the specified output path.
```
jskeet, line 45: But isn't this comment for the situation where we don't *have* an output path specified? It seems like a reasonable feature to keep.

jskeet (approved):
As usual, I've only skimmed this, but it seems reasonable.

src/Main.php
chrisdunelm: Maybe... OK, I won't remove it yet :)

PR #75, opened 2020-11-12
Refactor integration test

chrisdunelm:
Refactor only, no change in functionality

### DIFF
--- a/integration_tests/Invoker.php
+++ b/integration_tests/Invoker.php
@@ -1,0 +1,116 @@
+<?php
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


integration_tests/Invoker.php
```diff
@@ -0,0 +1,116 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+declare(strict_types=1);
+
+namespace Google\Generator\IntegrationTests;
+
+class Invoker
+{
+    public static function Invoke(string $protoPath)
```
jskeet, line 23: Do I want to know why this is `Invoke` but `SourceComparer` has a `compare` method (lower case)?

jskeet (approved):

integration_tests/Invoker.php
chrisdunelm: Oops! I'll fix the casing, just my mistake ;)

PR #77, opened 2020-11-12
Integration tests checks all generated files

chrisdunelm:


jskeet:
I won't be able to get to this one until next week I'm afraid.

amanda-tarafa:
On it

integration_tests/Main.php
```diff
@@ -31,17 +31,33 @@
...
+$excess = array_diff(array_keys($micro), array_keys($mono));
+$ok = count($excess) === 0 ? $ok : false;
+foreach ($excess as $excessPath) {
+    print("File mistakenly generated from micro-generator: '{$excessPath}'\n");
+}
+
+// Find incorrectly generated files.
+foreach (array_intersect(array_keys($mono), array_keys($micro)) as $path) {
+    print("Comparing: '{$path}':\n");
+    $sameContent = SourceComparer::compare($mono[$path], $micro[$path]);
```
amanda-tarafa, line 53: If the source comparer already prints something when it finds diffs, then ignore, else maybe print the name of the files that had diffs.

amanda-tarafa (approved):
LGTM

integration_tests/Main.php
chrisdunelm: Thanks for reviewing Amanda :)
The comparer currently prints out the diff if the files aren't the same.
I may well change this later, but for the moment it's probably OK as-is.

PR #80, opened 2020-11-13
Generate REST client config files.

chrisdunelm:
The micro-generator is now exactly matching monolith generation for a basic API.


amanda-tarafa (approved):
LGTM. yay!!!

PR #83, opened 2020-11-18
Update basic test to include scopes

chrisdunelm:
The monolith requires scopes to be defined in the service_config, it ignores scopes defined in the proto


amanda-tarafa (approved):
LGTM

PR #88, opened 2020-11-24
Conform proto docs handling between micro and monolith

chrisdunelm:
Including integration test


jskeet (commented):
Did you miss a file containing the expected generated output?

chrisdunelm:
No, I'm now testing using the integration testing. This runs both the (current) monolithic and micro generators and compares the outputs. The test passes if they are identical except for whitespace.

jskeet:
Ah, I see - okay, makes sense.

jskeet (approved):

PR #98, opened 2020-12-02
Test a first real API: language

chrisdunelm:
Changes required in output paths and namespace, as the 'version' part is treated specially.
Change required in doc handling, as some source locations are missing entirely, so handle this gracefully as empty documentation.


amanda-tarafa (approved):
LGTM

PR #99, opened 2020-12-04
Add LRO config and test/sample config from gapic-yaml file

chrisdunelm:


chrisdunelm:
This code will require a bit of refactoring later, but I'd like to get it in as-is today to allow people to try out the generator whilst I'm away next week.
All badness will be fixed upon my return ;)

amanda-tarafa (approved):
LGTM

PR #100, opened 2020-12-04
Enable grpc-service-config on cmd-line; generate correct retrys

chrisdunelm:
Reproduce the monolith behaviour of using different retry-config generation depending on whether the grpc-service-config is present or not.

This found a real genuine bug! Testing *does* work :)


amanda-tarafa (approved):
LGTM

PR #112, opened 2021-01-22
Fixes required for dataproc; and about to update monolith

chrisdunelm:
Dataproc revealed that the monolith has non-deterministic generation for placeholders; hence two integration tests currently removed. Next PR will be to update to latest monolith version which contains the fix; then to re-enable all tests

### DIFF
--- a/src/Generation/ResourcesGenerator.php
+++ b/src/Generation/ResourcesGenerator.php
@@ -33,6 +33,16 @@ use Google\Rpc\Code;
 
 class ResourcesGenerator
 {
+    // TODO(vNext): Remove this; only required for monolith compatibility.
+    private static function ensureDecimal(string $s): string
+    {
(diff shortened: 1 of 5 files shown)


src/Generation/ResourcesGenerator.php
```diff
@@ -33,6 +33,16 @@
 
 class ResourcesGenerator
 {
+    // TODO(vNext): Remove this; only required for monolith compatibility.
+    private static function ensureDecimal(string $s): string
+    {
+        if (strpos($s, '.') === false) {
```
jskeet, line 39: Would it make sense to use the conditional ?: operator here?

jskeet (approved):

src/Generation/ResourcesGenerator.php
chrisdunelm: Done :)

PR #113, opened 2021-01-22
Very minor refactor

chrisdunelm:
This should have been in the previous PR, but something went wrong...


chrisdunelm:
(Merging, as it's so trivial and green.)

PR #118, opened 2021-02-05
Update PHP deps

chrisdunelm:
protobuf appears to have a breaking change between 3.13.0 and 3.14.0, so fix affected code.


chrisdunelm:
Note that the `composer.lock` file is auto-generated, so doesn't really need checking.

amanda-tarafa (approved):
LGTM

PR #120, opened 2021-02-18
Bazel rules, with example of use

chrisdunelm:


chrisdunelm:
There are various repo references that are a little out of date.
These should be updated, but I'd rather update them in a separate PR.

PR #122, opened 2021-02-19
Add bazel build test; fix various issues

chrisdunelm:
Update versions of dependencies
Add missing gapic_generator_php_repositories invocation in bazel workspace.
Fix PHP source location, if not using pre-built package.
Fix file copying when performing composer install.
Fix php runner.
Correctly exit generator if EOF (^D) sent to stdin.


jskeet (approved):
Can't say I fully understand it, but it seems reasonable.

PR #125, opened 2021-02-23
Add support for specifying grpc config in bazel

chrisdunelm:
Allow config-file working directory to be specified on the cmd-line, which i  required when running from bazel.
Return generated files as an array rather than as a generator, to ease debugging.
Update prebuilt PHP to include bcmaths, as required by the proto JSON parser.

### DIFF
--- a/src/Main.php
+++ b/src/Main.php
@@ -33,7 +33,10 @@ function showUsageAndExit()
     print("Invalid arguments. Expect:\n");
     print("  --descriptor <path> The path to the proto descriptor file.\n");
     print("  --package <string> The proto package to generate.\n");
-    print("  --output <path> The output directory.\n");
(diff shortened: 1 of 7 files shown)


src/Main.php
```diff
@@ -44,8 +47,23 @@ function showUsageAndExit()
 $year = 2020;
 // $year = (int)date('Y');
 
-if ($argc === 1) {
-    // No args, probably being called from protoc.
+// When running as a protoc plugin, an optional root directory may be passed in to set the
+// root location of the side-loaded configuration files:
+// - grpc sevrice config json
```
jskeet, line 52:
```suggestion
// - grpc service config json
```

jskeet (approved):

jskeet:
(Accepting suggestions I've made via github always causes cla bot problems.)

PR #130, opened 2021-02-24
Fix building php from source

chrisdunelm:
Note that this is difficult to test; I've run this locally and in CI
with the pre-built php disabled to ensure that the built-from-sources
version of php works as expected.

Also update example to point to beta01 release of the generator rather than head.
And add comment explaining how php is executed within bazel.


amanda-tarafa (approved):
LGTM

PR #138, opened 2021-03-03
Add cloud functions to integration test

chrisdunelm:
And update googleapis to latest, to pull cloud-functions gapic-yaml fix.
And don't require paginated resources field to be field one, as some APIs don't follow the AIP here.
And add a YAML-fixer, as the PHP YAML parser doesn't support multiline string properly

### DIFF
--- a/integration_tests/Main.php
+++ b/integration_tests/Main.php
@@ -120,6 +120,21 @@ $ok = processDiff(Invoker::invoke(
     'googleapis/google/cloud/dialogflow/v2/dialogflow_v2.yaml',
     'googleapis/google/cloud/dialogflow/v2/dialogflow_grpc_service_config.json'
 )) ? $ok : false;
+// // TODO: googleapis/google/cloud/dialogflow/cx/v3/ has wrong capitalization somewhere, so monolith crashes
+// $ok = processDiff(Invoker::invoke(
+//     'googleapis/google/cloud/dialogflow/cx/v3/*.proto googleapis/google/cloud/common_resources.proto',
+//     'google.cloud.dialogflow.cx.v3',
+//     'googleapis/google/cloud/dialogflow/cx/v3/dialogflow_gapic.yaml',
+//     'googleapis/google/cloud/dialogflow/cx/v3/dialogflow_v3.yaml',
+//     'googleapis/google/cloud/dialogflow/cx/v3/dialogflow_grpc_service_config.json'
+// )) ? $ok : false;
+$ok = processDiff(Invoker::invoke(
+    'googleapis/google/cloud/functions/v1/*.proto googleapis/google/cloud/common_resources.proto',
+    'google.cloud.functions.v1',
+    'googleapis/google/cloud/functions/v1/functions_gapic.yaml',
+    'googleapis/google/cloud/functions/v1/cloudfunctions_v1.yaml',
+    'googleapis/google/cloud/functions/v1/functions_grpc_service_config.json'
+)) ? $ok : false;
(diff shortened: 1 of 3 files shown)


integration_tests/Main.php
```diff
@@ -120,6 +120,21 @@
     'googleapis/google/cloud/dialogflow/v2/dialogflow_v2.yaml',
     'googleapis/google/cloud/dialogflow/v2/dialogflow_grpc_service_config.json'
 )) ? $ok : false;
+// // TODO: googleapis/google/cloud/dialogflow/cx/v3/ has wrong capitalization somewhere, so monolith crashes
```
miraleung, line 123: Add a comment with a link or the specific failure case? 

Also FYI: For Java, the monolith did some other weird capitalization things we've had to special-case in the monolith. Not sure if this applies to PHP though. Cases that come to mind:
- "IAM" -> "Iam" but in general, FOOBar stays the same.
- 123abc becomes 123Abc. This was a legit bug because the monolith produced 123ABC and 123Abc in two different places (since fixed).

src/Generation/MethodDetails.php
```diff
@@ -115,7 +115,17 @@ private static function maybeCreatePaginated(ServiceDetails $svc, MethodDescript
         $pageSize = $inputMsg->desc->getFieldByName('page_size');
         $pageToken = $inputMsg->desc->getFieldByName('page_token');
         $nextPageToken = $outputMsg->desc->getFieldByName('next_page_token');
-        $resources = $outputMsg->desc->getFieldByNumber(1);
+        // Find the resoures field. Although AIP states that the resource field should be field number 1,
```
miraleung, line 118: Is this for the response message? If so, using the first repeated field handled all protos we came across in Java.

Curious: Will the repeated resource be needed for other parts of the generated code? If so (as in the case in Java), should this be extracted to a common parsing location?

src/Generation/MethodDetails.php
```diff
@@ -128,7 +138,7 @@ private static function maybeCreatePaginated(ServiceDetails $svc, MethodDescript
             if ($nextPageToken->isRepeated() || $nextPageToken->getType() !== GPBType::STRING) {
                 throw new \Exception("next_page_token field must be of type string.");
             }
-            if (!$resources->isRepeated() || ProtoHelpers::isMap($catalog, $resources)) {
+            if (!$resourceFieldValid) {
                 throw new \Exception("Item resources field must be a repeated field with field-number 1.");
```
miraleung, line 142: Maybe just have the "repeated" part, given the caveats above?

miraleung (approved):

src/Generation/MethodDetails.php
chrisdunelm: Oops, thanks, forgot to update this message. Done.

src/Generation/MethodDetails.php
chrisdunelm: I'll leave it as first-by-number-and-position for now, as I think that's mentioned in AIP and I think what C# does. This can always be updated if it needs to be more flexible.

So far, the repeated resource is not required anywhere else, but yes it should be exposed publicly on this class if that becomes the case. If I understand your question correctly, then this is the common parsing location, but at the moment the result isn't made available publicly (as it's not currently required).

integration_tests/Main.php
chrisdunelm: Yes, I suspect you're right that there's some pre-generator munging that happens for this API, which isn't reproduced in this test.

PR #144, opened 2021-03-04
fix(bazel): Fix Bazel build to work within this repo [ggphp]

miraleung:
`bazel build bazel_example:example_php_proto` within the repo works now.

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -14,5 +14,33 @@
 
 workspace(name = "gapic_generator_php")
 
-load("//repositories.bzl", "gapic_generator_php_repositories")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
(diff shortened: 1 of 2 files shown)


chrisdunelm:
I'm not sure building the example from this repo root is meant to work.
It's intended to be a stand-alone example in its own workspace, so you need to move into the `bazel_example` directory before invoking `bazel`.
However, I may have misunderstood the intent here, or not understood why it might be preferable this way; let's discuss before merging.

miraleung:
> I'm not sure building the example from this repo root is meant to work.
> It's intended to be a stand-alone example in its own workspace, so you need to move into the `bazel_example` directory before invoking `bazel`.
> However, I may have misunderstood the intent here, or not understood why it might be preferable this way; let's discuss before merging.

For posterity, marking as resolved after discussing offline.

PR #146, opened 2021-03-04
fix(bazel)!: remove unused package option from php_gapic_library [ggphp]

miraleung:


chrisdunelm (approved):
LGTM

chrisdunelm:
Has something gone wrong with CI tests here? If so, this is a fairly trivial change, so an admin-forced merge would seem fine to me :)

PR #152, opened 2021-03-05
Fix generated file layout; fix resource-names

chrisdunelm:
Fix file layout of generated code when there are files/namespaces beneath the top-level.
Fix the selection of which resource-name helpers to include in a generated client.
And add adsv6 to the integration test.

### DIFF
--- a/src/CodeGenerator.php
+++ b/src/CodeGenerator.php
@@ -61,7 +61,7 @@ class CodeGenerator
         $descSet->mergeFromString($descBytes);
         $fileDescs = Vector::new($descSet->getFile());
         $filesToGenerate = $fileDescs
-            ->filter(fn($x) => $x->getPackage() === $package)
+            ->filter(fn($x) => substr($x->getPackage(), 0, strlen($package)) === $package)
             ->map(fn($x) => $x->getName());
         yield from static::generate($fileDescs, $filesToGenerate, $licenseYear, $grpcServiceConfigJson, $gapicYaml, $serviceYaml);
     }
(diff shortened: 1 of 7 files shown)


src/CodeGenerator.php
```diff
@@ -127,8 +127,10 @@ private static function generatePackage(
         ?string $gapicYaml,
         ?string $serviceYaml
     ) {
-        $version = Helpers::nsVersion($namespace);
-        $version = is_null($version) ? '' : $version . '/';
+        $version = Helpers::nsVersionAndSuffixPath($namespace);
+        if ($version !== '') {
```
miraleung, line 131: Maybe add a brief comment on the intended behavior? After our discussion, I think I see what's going on, but a future reader may not.

src/Generation/ServiceDetails.php
```diff
@@ -139,22 +142,41 @@ public function __construct(
...
-            ->concat($typeRefResourceDefs)
-            ->concat($childTypeRefResourceDefs)
+        $msgsSeen = Set::new();
+        $gatherMsgResDefs = null;
+        $gatherMsgResDefs = function(DescriptorProto $msg, int $level) use(&$gatherMsgResDefs, &$msgsSeen, $catalog): Vector {
+            if ($msgsSeen[$msg->desc->getFullname()]) {
+                return Vector::new([]);
+            }
+            $msgsSeen = $msgsSeen->add($msg->desc->getFullname());
+            // Only top-level resource-defs are included; matches monolith behaviour.
```
miraleung, line 152: Add a TODO(vNext) or TODO(monoilith) here and below? (so that future readers can better identify this like the other notes in this codebase)

miraleung (approved):

src/CodeGenerator.php
chrisdunelm: done

src/Generation/ServiceDetails.php
chrisdunelm: done

PR #234, opened 2021-03-24
[ggp] fix(client): Handle optional and required request params (phase 2/2)

miraleung:
Fixes request parameter handling bugs compared to the monolith's behavior. Below is the file diff compared to the monolith-produced libraries.
- Container ([PR](https://github.com/googleapis/google-cloud-php/pull/3925))
- WebSecurityScanner ([PR](https://github.com/googleapis/google-cloud-php/pull/3916))
- Language ([PR](https://github.com/googleapis/google-cloud-php/pull/3917))
- BigQueryReservation ([PR](https://github.com/googleapis/google-cloud-php/pull/3918))

# Context

**Summary.** Analysis of the headers generated by the monolith for PHP client libraries, potential issues, and proposed fixes in the microgenerator.

**Issue.** The monolith populated request parameters in a manner that does not align with go/aip/4222. Request parameter values were produced based on unclear conditions, which led to inconsistent behavior with the same use case in other microgenerators. 

While the monolith usually generated request headers if and only if one HTTP binding was present, it would sometimes do so even when there were multiple bindings. Compare this to Java, where one request parameter is populated per HTTP binding. 

**Monolith Behaviour**
- [UpdateDocument](https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L70) has one HTTP binding, and [one request parameter](https://github.com/googleapis/google-cloud-php-firestore/blob/master/src/V1beta1/Gapic/FirestoreGapicClient.php#L698) generated.. This is correct, and aligns with other generators’ behavior (e.g. [Java](https://github.com/googleapis/java-firestore/blob/master/google-cloud-firestore/src/main/java/com/google/cloud/firestore/v1/stub/GrpcFirestoreStub.java#L300)).
- [UpdateNodePool](https://github.com/googleapis/googleapis/blob/master/google/container/v1/cluster_service.proto#L105) has one HTTP Put binding, and several POST bindings, but only the [first one](https://github.com/googleapis/google-cloud-php-container/blob/master/src/V1/Gapic/ClusterManagerGapicClient.php#L615) is accounted for.
  - For comparison, [Java](https://github.com/googleapis/java-container/blob/master/google-cloud-container/src/main/java/com/google/cloud/container/v1/stub/GrpcClusterManagerStub.java#L547) populates all the request parameters.
  - Similarly, [GetOperation](https://github.com/googleapis/googleapis/blob/master/google/container/v1/cluster_service.proto#L248) results in this [monolithic PHP](https://github.com/googleapis/google-cloud-php-container/blob/master/src/V1/Gapic/ClusterManagerGapicClient.php#L1421) behaviour, which differs from the microgenerated [Java client](https://github.com/googleapis/java-container/blob/master/google-cloud-container/src/main/java/com/google/cloud/container/v1/stub/GrpcClusterManagerStub.java#L710).
- [SubmitJob](https://github.com/googleapis/googleapis/blob/master/google/cloud/dataproc/v1/jobs.proto#L38) has two HTTP variables, but [no request parameter](https://github.com/googleapis/google-cloud-php-dataproc/blob/master/src/V1/Gapic/JobControllerGapicClient.php#L259) generated. 
  - For comparison, [Java](https://github.com/googleapis/java-dataproc/blob/master/google-cloud-dataproc/src/main/java/com/google/cloud/dataproc/v1/stub/GrpcJobControllerStub.java#L190) populates both request parameters.
- [CreateDocument](https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L196) has the HTTP variable `parent`, but the monolith [does not set](https://github.com/googleapis/google-cloud-php-firestore/blob/master/src/V1beta1/Gapic/FirestoreGapicClient.php#L642) this request parameter. (Java wrongly does not set this as well.)
- [CancelOperation](https://github.com/googleapis/googleapis/blob/master/google/container/v1/cluster_service.proto#L260) has the HTTP variable `name`. This is optionally passed into the corresponding PHP method, since the monolith was never upgraded to use the `method_signature` annotation. However, the monolith [still sets](https://github.com/googleapis/google-cloud-php-container/blob/master/src/V1/Gapic/ClusterManagerGapicClient.php#L1492) this request parameter, which can lead to a client-side PHP error if that optional argument is not provided.
	
  The monolith also does not set any of the other HTTP variables if they are present, in violation of [aip.dev/4222](aip.dev/4222). For comparison, [Java](https://github.com/googleapis/java-container/blob/master/google-cloud-container/src/main/java/com/google/cloud/container/v1/stub/GrpcClusterManagerStub.java#L710) sets these parameters.
- Given a proto like [this one](https://github.com/googleapis/gapic-generator-php/blob/master/tests/Unit/ProtoTests/RoutingHeaders/routing-headers.proto#L17), the monolith used to populate the `name` parameter. However, this could result in an error if that field is not set in the request. This violates the guidance in [aip.dev/4222](aip.dev/4222) that unset named parameters should not be included. The monolith currently generates the following, which will result in an error in PHP.
 
```
 $request = new SimpleRequest();
  if (isset($optionalArgs['name'])) {
      $request->setName($optionalArgs['name']);
  }
  $requestParams = new RequestParamsHeaderDescriptor([
      'name' => $request->getName()
  ]);
  // Truncated.
  return $this->startCall(
   'DeleteMethod', Response::class, $optionalArgs, $request)->wait();
```

**Proposal.** We can fix this in the PHP microgenerator by populating request parameters for every HTTP binding present. Since PHP constructs the request message using those method parameters, this adheres to [aip.dev/4222](aip.dev/4222), which states that any set variables in additional_bindings should be present.

This will require handling RPCs that have HTTP variables that are not marked as required fields on the corresponding request message. In this edge case, this parameter will not be set even if it is passed in via $optionalArgs to the PHP method. Since the generated surface would change significantly, we consider this out of scope for the microgenerator migration. 

The change for such a case should look like the following. We will not handle nested fields, since the generated PHP code for protobufs do not allow us to check whether a field is set or not. 

```
  $request = new SimpleRequest();
  $requestParamHeaders = [];

  $request->setRequiredArg($requiredArg);
  $requestParamHeaders[‘required_arg.foo’] = $requiredArg->getFoo();

  if (isset($optionalArgs['name'])) {
      $request->setName($optionalArgs['name']);
      $requestParamHeaders[‘name’] = $optionalArgs[‘name’];
  }

  $requestParams = new RequestParamsHeaderDescriptor($requestParamHeaders);
  $optionalArgs['headers'] = isset($optionalArgs['headers'])
      ? array_merge($requestParams->getHeader(), $optionalArgs['headers'])
      : $requestParams->getHeader();

  return $this->startCall(
   'DeleteMethod', Response::class, $optionalArgs, $request)->wait();
```

**Effects of this Change.** Extra attention will need to be given to APIs’ request parameters while migrating to the microgenerator, as the monolith-produced source cannot be used as a reliable reference.

**Alternatives**
- _Option 1_ - Attempt to recreate the monolith’s behavior. This is undesirable because it’s buggy, carries technical debt forward, and significant resources would be needed to parse the existing behavior.
- _Option 2_ - Generate request parameters for only the first binding present. This would not align with [aip.dev/4222](aip.dev/4222) or other generators, so the benefits of this partial implementation are unclear.
- _Option 3_ - Ignore the edge case handling above, and simply set request parameters for the required arguments present. This could potentially break clients like [Container](https://github.com/googleapis/google-cloud-php-container/blob/master/src/V1/Gapic/ClusterManagerGapicClient.php#L1492), which assume that optional fields’ parameters are being set.

### DIFF
--- a/src/Generation/GapicClientGenerator.php
+++ b/src/Generation/GapicClientGenerator.php
@@ -537,6 +537,7 @@ class GapicClientGenerator
         };
 
         $request = AST::var('request');
+        $requestParamHeaders = AST::var('requestParamHeaders');
         $required = $method->requiredFields->map(fn ($x) => AST::param(null, AST::var($x->camelName)));
         $optionalArgs = AST::param($this->ctx->type(Type::array()), AST::var('optionalArgs'), AST::array([]));
         $retrySettingsType = Type::fromName(RetrySettings::class);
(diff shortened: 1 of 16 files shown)


dwsupplee (approved):
The output looks good 👍

PR #268, opened 2021-04-23
[ggp] chore: Test unset optional value serialization

miraleung:
DO NOT SUBMIT
@vchudnov-g PTAL and confirm whether this test is sufficient.


validation/DefaultValuesValidator.php
```diff
@@ -0,0 +1,35 @@
...
+error_reporting(E_ALL);
+
+// Usage: php validation/DefaultValuesValidator.php
+
+$data = new ComplianceData();
+$data->setFString("fstring no presence")
+     ->setPString("P_string optional")
+     ->setPBool(False);
+$dataJson = $data->serializeToJsonString();
+assert($dataJson === ' {"fString":"fstring no presence","pString":"P_string optional","pBool":true}');
```
vchudnov-g, line 33: 1. in this assertion, `pBool` should be `false`, right?
2. Maybe do one more explicit zero-value check, that `setPString("")` returns `"pString": ""`
3. It looks like non-`optional` fields that have zero-values do not get transcoded? That should be fine (and not affect DIREGAPIC, only REGAPIC, and presumably beningly)
4. Are you planning on checking this file in as a test or otherwise? No need to, assuming we capture all this behavior in some other tests.

miraleung: No plans to check this in, this PR should be sufficient (and we can revive in the future).

dwsupplee (approved):

miraleung:
Merging so we can refer to source instead of this PR.

PR #276, opened 2021-05-05
fix(diregapic): pipe transport flag through Bazel, add Bazel REGAPIC integration test [ggp]

miraleung:


rules_php_gapic/php_gapic.bzl
```diff
@@ -61,15 +61,17 @@ def php_gapic_srcjar(
...
     if transport == None:
         transport = "grpc+rest"
     if transport == "grpc":
         fail("Error: gRPC-only PHP GAPIC libraries are not yet supported")
     if transport != "grpc+rest" and transport != "rest":
         fail("Error: Only 'grpc+rest' or 'rest' transports are supported")
-
-    # Generate the gapic_metadata.json file.
-    plugin_args = ["metadata"]
+    plugin_args.append("transport=%s" % transport)
```
vam-google, line 74: Why was this moved? It is not not a big deal, but it seems like the `transport` ifs block does not have impact on this value, so it is usually better to keep modifications to the same variable (`plugin_args` in this case) together (line 65 and line 74 in this case).

tests/Integration/BUILD.bazel
```diff
@@ -77,6 +78,27 @@ php_gapic_assembly_pkg(
...
+    deps = [
+        ":asset_rest_php_grpc",
+        ":asset_rest_php_proto",
+    ],
+)
+
+php_gapic_assembly_pkg(
+    name = "google-cloud-asset-rest-v1-php",
+    deps = [
+        ":asset_php_grpc",
```
vam-google, line 96: Should the rest client still depend on grpc stubs?

vam-google (approved):
LGTM (just a few minor comments).

rules_php_gapic/php_gapic.bzl
miraleung: The move allows us to group the `transport` setting logic together more tightly. Either has its merits, so I've moved it back.

PR #277, opened 2021-05-05
 feat: add Bazel flag for preview APIs [ggp]

miraleung:
This is needed to enable the `@experimental` tag on REGAPIC clients, since microgenerator's existing signals (the presence of `alpha` or `beta` in the package name) won't be sufficient, and gating on the transport type would also be a bad alternative.


vam-google (commented):
I feel like I misunderstood your original question about  `@experimental`. Was `@experimental` stuff present for DIREGAPIC clients in monolith implementation? I'm not sure that we need this change. This will also make the php generator surface inconsistent with the rest of the generators. 

Specifically, this PRs introduces the whole new concept of `preview` APIs, and that is visible on the rules surface (i.e. becomes an attribute of the generation). I don't think we had that before for the generators. 

For this specific case, gating `experimental` tag on the transport type does not seem that bad to me - it indicates that the whole feature of rest-only clients is experimental (which is true). I.e. here, the generated rest-only client is `experimental` but the compute API itself is not (compute is GA api, no alpha/beta stuff in the API's package name), that is why gating on the transport type seems appropriate. 

In other words, introducing extra flag (`preview`) and specifying it on generation time would technically mean that the compute API itself is experimental, which is not the case. We want to add `experimental` tag because the feature of regapic is experimental, not because compute the API is in preview phase.

For the API's-specific preview status we already have alpha/beta suffixes in their package names. Having also the `preview` flag will duplicate the information and will open the possibility for inconsistency between the two.

miraleung:
My original question on `@experimental` was essentially, "Should the GCE preview client have `@experimental` tags?" It appeared this would be needed, as per the 'alpha' or 'early testing' labelling requirement in the Private Preview email today. (And yes, the monolith-generated GCE client did have `@experimental` tags, but this was hardcoded for *all* PHP monolith-generated libraries.)

_> Specifically, this PRs introduces the whole new concept of preview APIs_
Happy to rename this flag or discuss alternate ways of adding the `@experimental` tag. Alternate approaches I had considered were:
- **Always add these tags on all REST-only transports.** This would be suboptimal once GCE goes GA but another API adopts REGAPIC clients in preview mode.
- **Hard-code on the `Compute` package name** (obviously suboptimal).
- **Use `alpha` / `beta` in the package name** - infeasible since GCE is at v1.

My original reasoning was that, while GCE itself is GA, the REGAPIC clients are not, so labelling that part of GCE as a preview should align with our interim state. WDYT?

_> we already have alpha/beta suffixes in their package names_
The versioning policy just changed, and I can send you some links to more internal resources on this.

vchudnov-g:
That email does not imply that we need to add preview labels to the code itself, but rather to people locating the GAPICs in package managers and in GitHub. So I think if we add the preview label to the name of the produced artifact, and have a a very clear message in relevant  READMEs (on GitHub and on the package manager), that should be sufficient. We can double-check with Han that this is sufficient.

====

That said, I'm not opposed to being able to label relevant sections of source code as preview (but let's be careful with YAGNI as per the above). Some thoughts:

One thing we need to be careful about is that for other APIs, we probably don't want to label the whole GAPIC as "preview". We might preview the REST transport part (though we haven't discussed it), but in general, unless configured otherwise, we will have a single GAPIC that will support both transports. Ideally, we would label only the REST specific parts as "preview" if necessary. [An alternative if we do have a REST preview for those other APIs might be to publish two GAPICs: the standard gRPC-only one as stable, and the gRPC+REST one as preview (and labeled as such). And then when we go GA, the gRPC+REST one becomes the stable one. This does seem like a lot of additional process, so we should discuss whether preview is needed for them, too.]

If it is indeed acceptable/possible to label only the REST part of the GAPIC as "preview", we could have the `preview` parameter be a string rather than a bool: `preview="rest"`. That way, generators know only that part of the GAPIC should be labeled as a preview.

miraleung:
Since the GCE client will not be in GA for a while, technically the `@experimental` tag should be applied until then. There isn't really a way to label the REST-only parts, since that does not affect the API surface.

That said, the other languages may not have this functionality in place, so we we don't want to add this label, that's fine as well.

miraleung:
Closing in favor of not putting these tags on the client librar.

PR #278, opened 2021-05-06
feat: support REGAPIC pagination, add compute_small goldens [ggp]

miraleung:
Adds proper support for REGAPIC pagination.


src/Generation/MethodDetails.php
```diff
@@ -131,104 +137,132 @@ private static function maybeCreatePaginated(ServiceDetails $svc, MethodDescript
...
         } else {
-            if ($pageSize->isRepeated() || $pageSize->getType() !== GPBType::INT32) {
-                throw new \Exception("page_size field must be of type int32.");
-            }
-            if ($pageToken->isRepeated() || $pageToken->getType() !== GPBType::STRING) {
-                throw new \Exception("page_token field must be of type string.");
-            }
-            if ($nextPageToken->isRepeated() || $nextPageToken->getType() !== GPBType::STRING) {
-                throw new \Exception("next_page_token field must be of type string.");
+            $isValidPageSize = $pageSize->getType() === GPBType::INT32;
```
vam-google, line 171: Potentially I'm reading it wrong (kind of hard to read the diff in github dif page), but it looks like this `else` block has changed the `if` statement it belonged to - it used to belong to `if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources))` and the rest of the if statements below were nested to it. Now it became the `else` for the `isValidPageSize` and the rest of the if statements are flat (not nested to this else block).  \

The whole sequence may still work as expected, but I wanted to make sure this was intentional. Was it? Even if it was, why was it done this way (basically the diff looks  much bigger than it actually is because of it).

tests/Integration/apis/compute_small/v1/compute_small.proto
```diff
@@ -0,0 +1,692 @@
+// Copyright 2020 Google LLC
```
vam-google, line 1: Looks like overall 14 new (mostly big) files were added to test the single max_results/page_size difference as a feature. This is a dangerous path (if we add so much stuff for every simple feature the tests will very quickly blow up - we ended up having exactly this problem for gapic-generator monolith). 

Can we just reuse the existing goldens for this? No need to introduce the whole compute_small thing in gapic-generator-php test infra. It is subjective, but maybe just not having the integration test for `max_results` is just ok  and would lead to a more maintainable code (as long as it is tested at least by unit tests).

miraleung: I've removed the `asset_rest` in favor of this, which will be the sole REGAPIC integration test (as I came across the `compute_small` protos only after writing #277.)

src/Generation/MethodDetails.php
miraleung: This diff is indeed hard to read. The original was something like this:

```
if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources)) {
    return null;
} else {
    if ($pageSize->isRepeated() || $pageSize->getType() !== GPBType::INT32) {
        throw new \Exception("page_size field must be of type int32.");
    }
   // More logic here.
   return new class // etc.
}
```

However, the `else` block is unnecessary and adds avoidable nesting. So this became:

```
if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources)) {
    return null;
}

if ($pageSize->isRepeated() || $pageSize->getType() !== GPBType::INT32) {
    throw new \Exception("page_size field must be of type int32.");
}
// More logic here.
return new class // etc.
```

Finally, the `$pageSize` check was altered to factor in REGAPIC-specific criteria:
```
if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources)) {
    return null;
}

$isValidPageSize = !$pageSize->isRepeated();
if ($isRestOnly) {
$isValidPageSize = $pageSize->getType() === GPBType::UINT32 
    || $pageSize->getType() === GPBType::INT32;
} else {
    $isValidPageSize = $pageSize->getType() === GPBType::INT32;
}
if (!$isValidPageSize) {
  // Throw exception.
}
// etc/
```

vchudnov-g (commented):
lgtm but please get approval from Vadym/Brent.

vam-google (approved):
LGTM

PR #292, opened 2021-05-18
fix(regapic): Remove PHP 5 workaround for list_ method renaming [ggp]

miraleung:
For GCE in conjunction with [the PHP 7 upgrade PR](https://github.com/googleapis/google-cloud-php/pull/3962/files).


bshaffer (approved):
Approved,  but depends on https://github.com/googleapis/google-cloud-php/pull/3962

miraleung:
Thanks @bshaffer, will regenerate the GCE client only after googleapis/google-cloud-php#3962 is checked in.

PR #307, opened 2021-06-10
fix(bazel): Remove monolith rule deps from the PHP µgen Bazel rules [ggp]

miraleung:


tests/Integration/BUILD.bazel
```diff
@@ -41,6 +41,108 @@ INTEGRATION_TEST_LIBRARIES = [
     target = "%s_php_gapic" % lib_name,
 ) for lib_name in INTEGRATION_TEST_LIBRARIES]
 
+###################################################
+# Temporary proto_with_info definitions for
+# monolith depdency removal.
+# Remove this section when the monolith rule deps are
+# out of googleapis.
+###################################################
+proto_library_with_info(
```
vam-google, line 50: Wow, I'm surprised that this works =) Please do not forget to remove this.

tests/Integration/goldens/logging/src/V2/Gapic/ConfigServiceV2GapicClient.php
```diff
@@ -78,7 +78,7 @@
  * ```
  * $configServiceV2Client = new ConfigServiceV2Client();
  * try {
- *     $formattedParent = $configServiceV2Client->organizationLocationName('[ORGANIZATION]', '[LOCATION]');
+ *     $formattedParent = $configServiceV2Client->locationName('[PROJECT]', '[LOCATION]');
```
vam-google, line 81: What is causing all these changes?

vam-google (commented):
LGTM, with several questions.

Also, technically speaking only monolith needs the `proto_library_with_info` rule at all (this one produces the `CustomProtoInfo` provider that is being mentioned in the error message). The fact that microgenerator started also depending on it was an unfortunate omission. 

Microgenerators should be able to depend directly on `proto_library` targets (producing ProtoInfo native provider), which is a built-in standard rule in bazel and does not have this chicken & egg problem.

tests/Integration/goldens/logging/src/V2/Gapic/ConfigServiceV2GapicClient.php
miraleung: I think it's a different version of the protos or a difference in the output of `proto_library_with_info`.

repositories.bzl
```diff
@@ -17,12 +17,20 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
...
 def gapic_generator_php_repositories():
+    _rules_gapic_version = "0.5.4"
+    maybe(
+        http_archive,
+        name = "rules_gapic",
+        strip_prefix = "rules_gapic-%s" % _rules_gapic_version,
+        urls = ["https://github.com/googleapis/rules_gapic/archive/v%s.tar.gz" % _rules_gapic_version],
+    )
+
+    # TODO: Remove this when the rule is gone from the monolith's switched_rules_by_language.
```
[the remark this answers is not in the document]
miraleung: Done.

miraleung:
> Microgenerators should be able to depend directly on `proto_library` targets (producing ProtoInfo native provider), which is a built-in standard rule in bazel and does not have this chicken & egg problem.

Thanks for the context, added a bug to the internal backlog.

PR #311, opened 2021-06-21
fix(bazel): Eradicate monolith Bazel deps from PHP µgen repo

miraleung:
Golden file updates are due to bumping the googleapis hash.


vam-google (approved):
LGTM

PR #316, opened 2021-07-09
fix: put useJwtAccessWithScope in credentialsConfig

bshaffer:
Context: The previous PR #309 adds `useJwtAccessWithScope` to the `$options` array but it needs to be in the subset `$options[credentialsConfig]` instead. 

Fortunately we haven't rolled out any clients with the previous PR.


miraleung (approved):
LGTM assuming the tests will be fixed.

PR #320, opened 2021-07-15
chore: Add baseline goldens for required oneofs [ggp]

miraleung:
Initial setup for the upcoming dev changes in go/php-gapic-required-oneofs. This PR makes future golden diffs easier to see.


.githooks/pre-commit
```diff
@@ -91,7 +91,8 @@ fi
 if [ $NUM_PHP_FILES_CHANGED -gt 0 ] || [ $NUM_UNIT_GOLDEN_FILES_CHANGED -gt 0 ]
 then
   echo_status "Checking unit tests..."
-  ./vendor/bin/phpunit --bootstrap tests/Unit/autoload.php tests/Unit
+  # TODO(miraleung: Remove the exclusion once oneofs are properly supported.
```
noahdietz, line 94: Close paren after your username.

noahdietz (approved):
Ah so these baselines are technically _wrong_, but once the required oneof feature is implemented, these baselines will be updated and we can review the diffs. Nice. LGTM

.githooks/pre-commit
miraleung: Done.

PR #325, opened 2021-07-21
feat(oneofs): initial scaffolding and wrapper class for required oneofs [ggp]

miraleung:
Sets up proto parsing for oneofs, and starts with an empty class. This will be merged into the `main/oneofs` branch, so  the empty class will not make it into prod.

In this PR, we’ll keep the existing naming convention “oneOf” in some places of the generator’s codebase (not the generated source). Since this is inconsistent with protobuf’s convention “one of” and hence is error-prone, we will rename all “[Oo]neOf” names to “[Oo]neof” in a future PR.

@dwsupplee  FYI. If the one of class should be located elsewhere, please LMK and I’ll update it in a future PR.


src/Generation/OneofWrapperGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+                if (sizeof($containingMessage->getOneofDecl()) === 0) {
+                    continue;
+                }
+
+                // Hazard zone: If the parsing logic in this codebase continues to work as intended,
+                // then the containing message's required fields should match up with those in $method.
+                // Ideally we'd use $method->requiredFields to construct the oneof descriptor, but any
+                // discrepancy could result in a field list mismatch against the containing message's
+                // fields in the oneof, when we generate the wrapper class. Either way, there will
+                // be error-prone risks, so we choose to align the werapper class with the containing
```
noahdietz, line 81: Nit: s/werapper/wrapper/

src/Generation/OneofWrapperGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+                $currOneofFieldNames = Vector::new([]);
+                // Vector of FieldDescriptors.
+                $currOneofFieldDescs = Vector::new([]);
+                // OneofDescriptorProto.
+                $oneofDescProto = $containingMessage->getOneofDecl()[$currOneofIndex];
+
+                // Collision between oneof names.
+                if ($uniqueOneofNames->contains($oneofDescProto->getName())) {
+                    throw new \Exception('Oneof ' . $oneofDescProto->getName() . ' in message '
+                        . $containingMessage->getName() . ' already exists elsewhere in another message');
```
noahdietz, line 93: Surely this can't be a problem for protobuf, so why is it a problem for us? My understanding is that this will throw an error when `message Foo` has a `oneof source` and `message Bar` has a `oneof source`. This will mostly certainly collide with some AIPs that define how to use `oneof` for standard RPCs i.e. Import/Export.

I think we need to prefix the oneof name (used for the class name, etc.) with the containing message name.

If I'm misunderstanding, please lmk.

src/Generation/OneofWrapperGenerator.php
miraleung: Done.

src/Generation/OneofWrapperGenerator.php
miraleung: Protobuf does not seem to generate oneof wrapper classes, but I see your point. Going with a subdirectory instead of a prefix, since some message names can get really long (e.g. Ads' [SuggestSmartCampaignBudgetOptionsRequest](https://github.com/googleads/google-ads-php/blob/master/src/Google/Ads/GoogleAds/V8/Services/SuggestSmartCampaignBudgetOptionsRequest.php#L25) and their oneof [`suggestion_data`](https://github.com/googleapis/googleapis/blob/67190ee7358af1341a3e15a666e507fafd7a9228/google/ads/googleads/v8/services/smart_campaign_suggest_service.proto#L57)).

noahdietz: > Protobuf does not seem to generate oneof wrapper classes

Right, I meant more that in protobuf, it is not a problem to have two messages with the same oneof name in each. 

> Going with a subdirectory instead of a prefix

Great idea.

noahdietz (approved):

PR #326, opened 2021-07-21
feat(oneofs): add private fields to the wrapper oneof class [ggp]

miraleung:
See API in go/php-gapic-required-oneofs. We're intentionally not generating typing here because that is supported only as of PHP 7.4+. We also don't add type hinting because that would be inconsistent with the rest of this client, and it would complicate the ongoing microgenerator migration for PHP clients.


noahdietz:
I'm guessing this branch needs to be rebased/merged to pick up the other PR that was merged.

noahdietz (approved):

PR #327, opened 2021-07-21
feat(oneofs): Add all methods for oneof wrapper class [ggp]

miraleung:


tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -67,4 +67,207 @@ class SupplementaryDataOneof
...
+
+    /**
+     * Sets this oneof to extra_description and updates its value.
+     *
+     * @param mixed $extraDescription The new value of this oneof.
+     */
+    public function setExtraDescription($extraDescription)
+    {
+        $this->extraDescription = $extraDescription;
+        $this->selectedOneofFieldName = extra_description;
```
noahdietz, line 79: Should this be `'extra_description'`/`"extra_description"`? Or is this a variable defined somewhere that I'm missing?

Same for the rest.

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -67,4 +67,207 @@ class SupplementaryDataOneof
...
+        $this->selectedOneofFieldName = extra_bool;
+        return $this;
+    }
+
+    /**
+     * Returns true if this oneof is set to the field extra_description.
+     */
+    public function isExtraDescription($extraDescription)
+    {
+        return $this->extraDescription === "extra_description";
```
noahdietz, line 160: I think `$this->extraDescription` should be `$this->selectedOneofFieldName`. Same elsewhere.

miraleung: Yes, done - thanks for the catch.

noahdietz (approved):

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -67,4 +67,207 @@ class SupplementaryDataOneof
...
+    /**
+     * Returns true if this oneof is set to the field extra_bool.
+     */
+    public function isExtraBool()
+    {
+        return $this->selectedOneofFieldName === 'extra_bool';
+    }
+
+    /**
+     * Returns $this->extraDescription if this oneof is set to the field
```
dwsupplee, line 212: WDYT? It's a small nit, but it seems a little out of place to reference the PHP syntax in the description here.

```suggestion
     * If the oneof field field extra_description is set it returns its value, null otherwise.
```

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -67,4 +67,207 @@ class SupplementaryDataOneof
...
+     * extra_description, null otherwise.
+     */
+    public function getExtraDescription()
+    {
+        return $this->isExtraDescription() ? $this->extraDescription : null;
```
dwsupplee, line 217: If you prefer, there is a shorthand operator that can be used:

```suggestion
        return $this->isExtraDescription() ?: $this->extraDescription;
```

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
miraleung: Will change `$this->extraDescription` in #333, but leaving the rest as-is. Since that field can still be internally set to a value even if it is not the "selected" oneof, the existing description seems more accurate.

PR #329, opened 2021-07-23
fix(oneofs): fix oneof wrapper field access [ggp]

miraleung:


noahdietz (approved):
Oops! I think I should've caught that in review, my bad.

dwsupplee (approved):

PR #333, opened 2021-07-28
feat(oneofs): add oneofs usage to sample code and polish generated PhpDocs [ggp]

miraleung:


src/Generation/GapicClientGenerator.php
```diff
@@ -755,15 +755,17 @@ private function rpcMethod(MethodDetails $method): PhpClassMember
...
                         PhpDoc::param(
                             $param,
                             PhpDoc::preFormattedText(
                                 !$field->isOneOf
                                     ? $field->docLines->concat($docExtra($field))
-                                    : Vector::new(['Maps to the required proto oneof '
-                                        . $field->containingMessage->getOneofDecl()[$field->oneOfIndex]->getName()
-                                        . '.'
+                                    : Vector::new([
+                                        'Maps to the required proto oneof ' . $field->getOneofDesc()->getName() . '.'
```
fiboknacky, line 768: As it's a wrapper class now, could you change this to something else that is not "Maps"?
I think it's misleading to code readers.

miraleung: Done.

src/Generation/GapicClientExamplesGenerator.php
```diff
@@ -93,7 +93,15 @@ private function initCallVars(MethodDetails $method): array
...
-                        $initCode = AST::assign($var, $f->exampleValue($this->ctx));
+                        if ($f->isOneOf) {
+                            $initCode = AST::assign(AST::var(Helpers::toCamelCase($f->containingMessage->getOneofDecl()[$f->oneOfIndex]->getName())),
+                                AST::call(
+                                    AST::new($this->ctx->type($f->toOneofWrapperType($method->serviceDetails->namespace)))(),
+                                    AST::method("set" . Helpers::toUpperCamelCase($f->camelName)))($f->exampleValue($this->ctx))
+                            );
+                        } else {
+                            $initCode = AST::assign($var, $f->exampleValue($this->ctx));
+                        }
```
noahdietz, line 104: Nit: Is this just a code style thing to move the "default case" into the `else` rather than have it always execute prior to the "special case" `if ($f->isOneOf)`?

noahdietz (approved):

src/Ast/AST.php
```diff
@@ -435,7 +435,12 @@ public function toCode(): string
                 if (is_null($this->callee)) {
                     return static::toPhp($this->obj) . "({$args})";
                 } else {
-                    return static::toPhp($this->obj) . static::deref($this->obj) . static::toPhp($this->callee) . "({$args})";
+                    // Handle calling a function directly on a constructor.
+                    $objCode = static::toPhp($this->obj);
+                    if (substr($objCode, 0, 4) === 'new ') {
```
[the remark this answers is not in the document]
miraleung: The string length of the literal `'new '`. Added a comment.

src/Generation/GapicClientExamplesGenerator.php
miraleung: Yes. Similar patterns elsewhere made code parsing easier for me, but I don't have a strong preference.

noahdietz: Ok! That's fine with me, especially for consistency.

My personal preference is to keep "basic case" code executing before the conditionals for the "special" cases, if that code has no undesirable side effects, like this which is just assigning a value to a variable. That keeps the main execution of the code left justified and the special cases indented.

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -26,7 +26,9 @@
 
 use Testing\BasicOneof\PartOfRequestA;
 
-/** Wrapper class for the oneof supplementary_data defined in message  */
+/** Wrapper class for the oneof supplementary_data defined in message Request.
+Only one item should be set on an instance of this class.
+If multiple items are set on the instance, the last one is used. */
```
dwsupplee, line 31:
```suggestion
/** 
 * Wrapper class for the oneof supplementary_data defined in message Request.
 * Only one item should be set on an instance of this class.
 * If multiple items are set on the instance, the last one is used. 
 */
```

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -98,6 +104,8 @@ public function setExtraSummary($extraSummary)
      * Sets this oneof to extra_request and updates its value.
      *
      * @param PartOfRequestA $extraRequest The new value of this oneof.
+     *
+     * @return SupplementaryDataOneof The modified object
      */
     public function setExtraRequest($extraRequest)
```
dwsupplee, line 110: Would we be able to typehint in cases like this where it is feasible?

tests/Unit/ProtoTests/BasicOneof/out/src/Gapic/BasicOneofGapicClient.php
```diff
@@ -172,21 +167,15 @@ public function __construct(array $options = [])
...
+     *     $supplementaryData = (new SupplementaryDataOneof())->setExtraDescription('extra_description');
+     *     $response = $basicOneofClient->aMethod($extraDescription);
      * } finally {
      *     $basicOneofClient->close();
      * }
      * ```
      *
-     * @param string $supplementaryData Maps to the required proto oneof supplementary_data.
-     * @param array  $optionalArgs      {
+     * @param SupplementaryDataOneof $supplementaryData An instance of the wrapper class for the required proto oneof supplementary_data.
```
dwsupplee, line 177: Can we typehint the signature?

dwsupplee (commented):
These thoughts are strictly around the generated output, would be happy to look over the generator code shortly as well.

Two thoughts I'd love to get your opinion on @miraleung:

1. What are your thoughts on replacing the `is*\get*` methods on the OneOf class with a simple `getOneOfValue`? This would reduce boilerplate in both the client/OneOf implementation, as I can't think of too many cases where a user would need to utilize the specific is/get methods.
2. Is it safe to use the `Request` directory to house the generated OneOf class? Since it is relatively generic, it seems possible it could conflict with an existing or future service. Also, it seems like an odd man out since protobuf generates request based classes and doesn't segregate them into a special directory currently. I personally like the idea of having the OneOf class live and be namespaced right next to the generated protobuf classes so they act/appear of the same kin.

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
miraleung: Since we don't typehint in other GAPIC methods (and I'm not sure if our supported versions of PHP for clients would support this), would it be odd or breaking to add it only here? For these reasons, I'd lean towards not typehinting ATM, and doing a major upgrade for all generated methods at some future point.

miraleung:
> These thoughts are strictly around the generated output, would be happy to look over the generator code shortly as well.
> 
> Two thoughts I'd love to get your opinion on @miraleung:
> 
> 1. What are your thoughts on replacing the `is*\get*` methods on the OneOf class with a simple `getOneOfValue`? This would reduce boilerplate in both the client/OneOf implementation, as I can't think of too many cases where a user would need to utilize the specific is/get methods.

The rationale for using the individual methods is covered in the design doc. I'd prefer to use the is/get methods to avoid complications with mixing up types.

> 2. Is it safe to use the `Request` directory to house the generated OneOf class? Since it is relatively generic, it seems possible it could conflict with an existing or future service. Also, it seems like an odd man out since protobuf generates request based classes and doesn't segregate them into a special directory currently. I personally like the idea of having the OneOf class live and be namespaced right next to the generated protobuf classes so they act/appear of the same kin.

The `Request` here is actually the name of the request message. If we named this "FooRequest", that would be the name of the directory as well (as per existing protobuf conventions when generating extra things for a message type). PTAL as well at Ads' PR for an example of how this works in practice (see email thread).

tests/Unit/ProtoTests/BasicOneof/out/src/Gapic/BasicOneofGapicClient.php
miraleung: Ditto.

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
dwsupplee: Typehinting against objects is supported in 5.5+. I hadn't realized we don't typehint elsewhere, good point and I agree - let's do it all at once later.

dwsupplee:
Thanks for the additional context/thoughts. Seems reasonable enough to me.

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
miraleung: Done (used a fix in the generator).

tests/Unit/ProtoTests/BasicOneof/out/src/Gapic/BasicOneofGapicClient.php
```diff
@@ -172,21 +167,15 @@ public function __construct(array $options = [])
...
-     *     $extraDescription = 'extra_description';
-     *     $extraSummary = 'extra_summary';
-     *     $extraRequest = new PartOfRequestA();
-     *     $extraIndex = 0;
-     *     $extraDouble = 0.0;
-     *     $extraFloat = 0.0;
-     *     $extraBool = false;
-     *     $response = $basicOneofClient->aMethod($extraDescription, $extraSummary, $extraRequest, $extraIndex, $extraDouble, $extraFloat, $extraBool);
+     *     $supplementaryData = (new SupplementaryDataOneof())->setExtraDescription('extra_description');
+     *     $response = $basicOneofClient->aMethod($extraDescription);
```
dwsupplee, line 171:
```suggestion
     *     $response = $basicOneofClient->aMethod($supplementaryData);
```

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
```diff
@@ -213,6 +245,8 @@ public function isExtraBool()
     /**
      * Returns $this->extraDescription if this oneof is set to the field
      * extra_description, null otherwise.
+     *
+     * @return string
```
dwsupplee, line 249:
```suggestion
     * @return string|null
```

tests/Unit/ProtoTests/BasicOneof/out/src/Gapic/BasicOneofGapicClient.php
miraleung: Done.

tests/Unit/ProtoTests/BasicOneof/out/src/Request/SupplementaryDataOneof.php
miraleung: Done.

miraleung:
Merging with @dwsupplee's LGTM offline.

PR #348, opened 2021-07-30
chore(engx): Remove monolith comparison tests [ggp]

miraleung:


noahdietz (approved):

.gitmodules
```diff
@@ -1,12 +1,6 @@
 [submodule "protobuf"]
 	path = protobuf
 	url = https://github.com/protocolbuffers/protobuf.git
-[submodule "googleapis"]
```
noahdietz, line 4: Should this be removed?

miraleung: Done, added this back.

PR #355, opened 2021-08-05
chore: only use tools/protoc if enabled

noahdietz:
The `tools/protoc` binary is not executable on mac os, my guess is because it is a linux binary. So instead of always using that, all scripts/execs default to using `protoc` (local installation in `PATH`) unless `USE_TOOLS_PROTOC` is set.

Also fixes a function that was non-static but called in a static context.

### DIFF
--- a/tests/Tools/ProtoLoader.php
+++ b/tests/Tools/ProtoLoader.php
@@ -39,7 +39,7 @@ class ProtoLoader
         $cwd = getcwd();
         $protoc = "protoc";
         $useToolsProtoc = getenv("USE_TOOLS_PROTOC");
-        if ($useToolsProtoc && $useToolsProtoc == "true") {
+        if (filter_var($useToolsProtoc, FILTER_VALIDATE_BOOLEAN)) {
             $protoc = "{$cwd}/tools/protoc";
         }
         $descRes = tmpfile();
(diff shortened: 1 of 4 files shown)


miraleung (approved):

tests/Tools/ProtoLoader.php
```diff
@@ -37,7 +37,10 @@ public static function loadDescriptorBytes(string $protoPath): string
         // Set up required file locations and create tmp output file for protoc invocation.
         // Assumes test are executed from within the repo root directory.
         $cwd = getcwd();
-        $protoc = "{$cwd}/tools/protoc";
+        $protoc = "protoc";
+        if (getenv("USE_TOOLS_PROTOC")) {
```
bshaffer, line 41: if `export USE_TOOLS_PROTOC=`, then this will be false but the bash conditional will be true. to ensure there's no discrepancy, I'd check for the string `true` in both cases, or the string `1`, but be specific.

bshaffer (approved):
One suggestion, otherwise LGTM

tests/Tools/ProtoLoader.php
noahdietz: Done! Switched it to compare to `'true'`.

dwsupplee: Another good option is `filter_var($var, FILTER_VALIDATE_BOOLEAN)`. This will give you an actual boolean to work with regardless of casing, etc.

PR #375, opened 2021-09-01
chore: add custom LRO proto configs

noahdietz:
Adds the generated type for `google/cloud/extended_operations.proto` and declares new `CustomOption`s by extension number for each annotation.


noahdietz:
@bshaffer could I get a quick review? This is just adding the PHP bindings/constants for the new extended_operations.proto we will need for diregapic LRO development

bshaffer:
Is this where these should go? Wouldn't common-protos-php be more appropriate?

noahdietz:
> Is this where these should go? Wouldn't common-protos-php be more appropriate?

Well, the CustomOptions thing was already there for all of the other annotations, so I just added to that set of constants. And the Enum  `OperationResponseMapping` I figured could go there because the LRO OperationInfo message (used in the annotation) is in here already (as well as other common proto types that seem to be in common-proto-php already).

noahdietz:
Interestingly, this repo doesn't have a dep on `common-proto-php` atm!

noahdietz:
So there are a number of things that `common-protos-php` doesn't have that are kind of specific to the generator - GAPIC Metadata schema, gRPC service config for retry/timeout config at generation time, some Showcase proto bindings, the protobuf CodeGenerator(Request|Response) types, etc. I think `common-protos-php` makes sense as a dependency for client libraries and some tools, but not necessarily for the generator. These protos in particular are very specific to generator use.

src/Utils/CustomOptions.php
```diff
@@ -29,4 +29,12 @@ class CustomOptions
     public const GOOGLE_API_RESOURCEDEFINITION = 1053;
 
     public const GOOGLE_API_FIELDBEHAVIOR_REQUIRED = 2;
+
+    // Protobuf extension numbers from https://github.com/googleapis/googleapis/blob/master/google/cloud/extended_operations.proto.
+    public const GOOGLE_CLOUD_OPERATION_FIELD = 1149;
```
bshaffer, line 34: It's interesting that we copy these values over instead of generating new `google.protobuf.FieldOptions` and `google.protobuf.MethodOptions` objects, but I see that this approach is simpler and so if the actual message classes aren't needed, it's probably just as good to leave them out.

bshaffer (approved):
Since this doesn't have a dep on `common-protos-php`, it makes sense to me to just merge the annotation changes here, and if we need to we can put them in `common-protos-php` also (which is the case right now for the `Google\LongRunning` protos.

PR #403, opened 2021-10-21
feat: add RoutingRule explicit header injection support

noahdietz:
Add support for the [`google.api.routing`] annotation. In short, this annotation allows API producers to _explicitly_ define those request fields should be include in request headers/metadata. Furthermore, it enables API producers to define simple segment matchers (primarily for resource name values) that are generated into regular expressions in the client that extract segments of a field's value to be sent in the header. Finally, it enables API producers to specify a key other than the field name to use in the header key-value-pair. In relation to the _implicit_ headers (extraced from `google.api.http` annotations), the `google.api.routing` annotation takes precedence. Please read the annotation documentation for more information on the annotation resolution logic.

Notes:
* all examples from [`google.api.routing`] are covered in the proto unit test
* PHP GAPICS have REQUIRED fields as positional arguments and OPTIONAL as array key/value pairs so the generated code accesses them slightly differently
* With multi-level nested fields, there is always the chance that, unless all are REQUIRED, one of the messages in the chain might be `null` and an exception thrown. That is why today, the _implicit_ headers don't support multi-level nested fields
* headers are already encoded by gax-php

[`google.api.routing`]: https://github.com/googleapis/googleapis/blob/master/google/api/routing.proto


noahdietz:
@viacheslav-rostovtsev if you have some time an want to review this work, please let me know. Otherwise, feel free to remove yourself from the reviewers.

noahdietz:
@bshaffer PTAL when you have a chance. Take a look at the code generated for the test package as an example of how this will work. LMK if you need more documentation.

bshaffer (approved):

src/Utils/ProtoHelpers.php
```diff
@@ -76,6 +78,150 @@ public static function isMap(ProtoCatalog $catalog, FieldDescriptor $desc): bool
         }
     }
 
+    /**
+     * Processes the RoutingParameters for generation and groups these processed configs by the `field`
+     * declared.
+     *
+     * @param ProtoCatalog $catalog The proto catalog.
+     * @param DescriptorProto $msg The request message.
+     * @param RoutingRule $routingRule The RoutinRule annotation to process.
```
alicejli, line 87: nit:
```suggestion
     * @param RoutingRule $routingRule The RoutingRule annotation to process.
```

alicejli (approved):
tiny nit - otherwise looks good. Thanks for your hard work on this!

src/Utils/ProtoHelpers.php
bshaffer: This is just some southern-talk, y'all

PR #414, opened 2021-11-16
feat: include server streaming in rest descriptor

noahdietz:
Enables generation of a REST descriptor for server-stream RPCs in order to support REGAPIC streaming.


src/Generation/ResourcesGenerator.php
```diff
@@ -210,7 +210,7 @@ public static function generateRestConfig(ServiceDetails $serviceDetails, Servic
     private static function compileRestConfigInterfaces(ServiceDetails $serviceDetails, ServiceYamlConfig $serviceYamlConfig)
     {
         return $serviceDetails->methods
-            ->filter(fn ($method) => !is_null($method->httpRule) && !$method->isStreaming() && !$method->isMixin())
+            ->filter(fn ($method) => !is_null($method->httpRule) && ($method->isServerStreaming() || !$method->isStreaming()) && !$method->isMixin())
```
bshaffer, line 213: Why is this necessary, when `$this->methodType === static::SERVER_STREAMING` is included in the function `isStreaming`?

src/Generation/MethodDetails.php
```diff
@@ -536,6 +536,11 @@ public function isStreaming(): bool
             $this->methodType === static::CLIENT_STREAMING;
     }
 
+    public function isServerStreaming(): bool
+    {
+        return $this->methodType === static::SERVER_STREAMING;
+    }
```
bshaffer, line 542: I think this function may be unnecessary...

src/Generation/ResourcesGenerator.php
noahdietz: We _want_ to enable rest config generation `SERVER_STREAMING` but not `CLIENT_STREAMING` or `BIDIRECTIONAL_STREAMING` which `isStreaming` returns `true` for.

src/Generation/MethodDetails.php
noahdietz: I explained in the previous comment relating to this. We want to generate a rest descriptor config for server streaming, but _not_ the other two types of streaming (because we haven't been).

src/Generation/ResourcesGenerator.php
bshaffer: Oops sorry, I missed the good ol `!` at the start of `$this->isStreaming()`

bshaffer (approved):

PR #423, opened 2021-12-07
feat(diregapic): generate string constants for enum names

noahdietz:
For DIREGAPIC APIs, this generates a class for each enum containing nothing but string constants for each of the enum's value names. The classes are namespaced within the GAPIC according to their relative scope in the proto file (top-level enum or enum nested in a message or nested message). See go/actools-gce-gapic-enum-string-impl for more info.

Also updates the version of composer checked into the repo for bazel.


tests/Unit/ProtoTests/CustomLro/out/src/Gapic/CustomOperationResponse/Status.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/*
+ * GENERATED CODE WARNING
+ * Generated by gapic-generator-php from the file
+ * https://github.com/googleapis/googleapis/blob/master/tests/Unit/ProtoTests/CustomLro/custom_lro.proto
+ * Updates to the above are reflected here through a refresh process.
+ */
+
+namespace Testing\CustomLro\Gapic\CustomOperationResponse;
```
bshaffer, line 25: This looks good! my one suggestion: change the namespace to be `Enums` instead of `Gapic`

noahdietz: Great idea. Done!

noahdietz:
@bshaffer PTAL!

src/CodeGenerator.php
```diff
@@ -305,4 +318,39 @@ private static function generateServices(
             }
         }
     }
+
+    private static function generateEnumConstants(Map $filesByPackage, ProtoCatalog $catalog, int $licenseYear)
```
alicejli, line 322: Could you rename $filesByPackage to $inputFilesByPackage or something similar?

alicejli (approved):
Overall looks good! Something to consider in the future is to create a function for making something a fully qualified name (e.g. adding the `.` in front of a string) for easier readability. Thank you for your hard work on this!

src/CodeGenerator.php
noahdietz: Good idea. Done!

noahdietz:
> Something to consider in the future is to create a function for making something a fully qualified name (e.g. adding the `.` in front of a string) for easier readability.

Great call. Let's do this in a subsequent PR across the repo. I've filed #427 to track this work. Thanks for the suggestion!

bshaffer (approved):

PR #424, opened 2021-12-07
fix(diregapic): use camelCase field name in custom op test

noahdietz:
Accidentally used plain field `name` which for field names with `_` in them, isn't the proper variable name. Using `camelCase` is the right thing.


bshaffer (approved):

bshaffer (commented):
Only one question - shouldn't there be a test to get updated from this?

noahdietz:
I'm going to circle back with a test. I tested locally against compute already.

PR #428, opened 2021-12-10
fix(bazel): update composer to version 2.1.14 from 2.1.5

alicejli:
This updates the copy of composer used by the PHP client library generator to the latest stable version, by using `rules_php_gapic/resources/composer.phar self-update --stable`. This updates composer to version `2.1.14` from `2.1.5`.

This should fix the issue where some users are unable to build their PHP client libraries due to the error:

`STDERR: Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled.
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.`


noahdietz:
Let's do a couple of things to help document this situation:
- Please add a section to the readme for composer self-update (explicitly _no_ snapshot versions, we want stable channels)
- Let's make sure we document which version is currently in use - probably in the same readme
- I would like to log the composer version used at run time - let's see if we can add a `composer --version` step to the Bazel  rule just before `composer install` is called

All of this can come in a follow up PR but please make sure to 1. use a stable channel in this update 2. add the version to the commit message/PR message so we know what this PR changed it to.

rules_php_gapic/resources/readme.md
```diff
@@ -64,3 +64,13 @@ The recommended (tested) way of doing it is as follows:
     tar -zchpf php-7.4.15_linux_x86_64.tar.gz php-7.4.15/bin php-7.4.15/lib
     ```
 11. Copy the created archive to the `rules_gapic/php/resources` folder in this repository and post a PR.
+
+## Updating composer
+
+When generating client libraries, this generator is bundled together with a version of composer. Occasionally composer needs to be updated in order for bazel to build successfully. You may see a somewhat misleading error when trying to build with bazel with an out-of-date version of composer:
+
+```
+STDERR: Composer is operating significantly slower than normal because you do not have the PHP curl extension enabled. No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
```
noahdietz, line 73: I'm not sure that this error is correlated with an outdated version, so lets not mention it here. Furthermore, these docs are relevant even when there is nothing wrong, nut we just want to update composer.

DEVELOPMENT.md
```diff
@@ -141,4 +141,8 @@ From the `googleapis` submodule, generate the PHP Protobuf bindings with
 
 Note: If the newly generated file belongs to a new package, ensure that it is
 in scope of the existing autoload rules in the [composer.json](./composer.json),
-adding it if it's not.
\ No newline at end of file
+adding it if it's not.
+
+## How to update composer
```
noahdietz, line 146:
```suggestion
## How to update composer for use in Bazel
```

noahdietz (approved):
LGTM save for two nits

DEVELOPMENT.md
alicejli: ack + done. Thanks!

PR #439, opened 2022-01-20
feat: add manual fix for bc-breaking changes

bshaffer:


noahdietz:
I think now the test is failing because of the tricky `autoload` functionality https://github.com/googleapis/gapic-generator-php/blob/main/tests/Unit/autoload.php that resolves proto messages and field values with dumbies.

bshaffer:
@noahdietz any idea what the deal is with the failing bazel integration tests?

noahdietz:
> @noahdietz any idea what the deal is with the failing bazel integration tests?

You'll need to run `bazel run tests/Integration:asset_update` (`logging_update` and `videointelligence_update`) to update those golden files since this change affects their generated code.

Sorry for all of the toil.

bshaffer:
@noahdietz since we have the Integration tests already, which cover the same thing, should I just get rid of the tests i added to `tests/Unit`?

noahdietz:
> @noahdietz since we have the Integration tests already, which cover the same thing, should I just get rid of the tests i added to `tests/Unit`?

Yeah, that is probably fine.

PR #465, opened 2022-02-23
fix: update PHP prebuilt binary, version 7.4.15

alexander-fenster:
Using `googleapis` Docker image `20220222`, now with curl enabled. Replaces #462.


noahdietz (approved):
This will ensure that the binary works inside our specific docker image, but not necessarily when run from a linux workstation or other docker images (unless OwlBot, etc. us the same image). I wonder if we need to include more cached binaries to improve performance in those use cases. Speeding up the kokoro runs is definitely most important though.

noahdietz:
> I actually checked that it works on a Linux workstation (in its current state).

Awesome! Thank you for doing that.

PR #476, opened 2022-04-19
fix: ignore synthetic oneof from proto3_optional

### LINKED ISSUE #475: Ignore field_behavior for synthetic oneofs created by proto3_optional
The `proto3_optional` feature results in fields labeled as `optional` being presented as a oneof to plugins by the compiler, even though the field isn't a oneof. This is a synthetic oneof. This means that if a `proto3_optional` field is also marked with `google.api.field_behavior = REQUIRED`, the required oneof wrappers are generated for a field that isn't actually a oneof (just presented as one by the protobuf compiler).

Generation of the required oneof wrapper types should ignore synthetic oneofs i.e. for those fields that are marked `proto3_optional`.

noahdietz:
Fixes #475

### DIFF
--- a/src/Generation/OneofWrapperGenerator.php
+++ b/src/Generation/OneofWrapperGenerator.php
@@ -59,13 +59,13 @@ class OneofWrapperGenerator
         // Use a vector instead of a set, since adding contains() to Set would have a similar
         // performance cost not lower than calling Vector::contains() directly.
         $oneofs = Map::new([]);
-        $oneofContainingMessageNames = Vector::new([]);
         foreach ($this->serviceDetails->methods as $method) {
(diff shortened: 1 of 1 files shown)


src/Generation/OneofWrapperGenerator.php
```diff
@@ -59,13 +59,13 @@ private function generateImpl(): Vector
         // Use a vector instead of a set, since adding contains() to Set would have a similar
         // performance cost not lower than calling Vector::contains() directly.
         $oneofs = Map::new([]);
-        $oneofContainingMessageNames = Vector::new([]);
```
noahdietz, line 62: This was an unused variable

bshaffer (approved):
Looks good but... Can we add a test?

PR #478, opened 2022-04-20
fix: better fix for ignoring required synthetic oneofs

noahdietz:
This is a better fix for #475 than #476, ignoring the synthetic `oneof` at the FieldDetails level rather than everywhere the field is checked as a `oneof`.

Also allow use of PHP 8.x with the generator.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         }
     },
     "require": {
-        "php": "^7.4",
+        "php": "^7.4 || ^8.0",
         "google/protobuf": "^3.18",
         "friendsofphp/php-cs-fixer": "^2.16",
         "symfony/yaml": "^5.2"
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
         }
     },
     "require": {
-        "php": "^7.4",
+        "php": "^7.4 || ^8.0",
```
bshaffer, line 29: Did we get a request to allow for 8.0? Should we bump this to an 8.0 requirement at some point?

bshaffer (approved):

composer.json
noahdietz: > Did we get a request to allow for 8.0?

Not explicitly, but since we are now supporting it for clients, I might want to be able to run tests in multiple versions.

> Should we bump this to an 8.0 requirement at some point?

Yeah, possibly. This will take a little coordination as we will need to ensure the bazel build pipeline has the proper PHP version(s).

PR #483, opened 2022-05-27
feat: introduce snippet gen support

dwsupplee:


tests/Unit/ProtoTests/BasicBidiStreaming/out/samples/BasicBidiStreamingClient/method_bidi.php
```diff
@@ -0,0 +1,71 @@
...
+        $requests = [
+            $request,
+        ];
+        $stream = $basicBidiStreamingClient->methodBidi();
+        $stream->writeAll($requests);
+        foreach ($stream->closeWriteAndReadAll() as $element) {
+            printf('Element data: %s' . PHP_EOL, $element->serializeToJsonString());
+        }
+
+
```
bshaffer, line 52: nit: extra whitespace here

tests/Unit/ProtoTests/BasicBidiStreaming/out/samples/BasicBidiStreamingClient/method_bidi.php
```diff
@@ -0,0 +1,71 @@
...
+ * This file was automatically generated - do not edit!
+ */
+
+require_once __DIR__ . '../../../vendor/autoload.php';
+
+// [START bidi_generated_BasicBidiStreaming_MethodBidi_sync]
+use Google\ApiCore\ApiException;
+use Testing\BasicBidiStreaming\BasicBidiStreamingClient;
+
+
```
bshaffer, line 29: nit: should just one newline between `use` and the top of `phpdoc`

tests/Unit/ProtoTests/BasicBidiStreaming/out/samples/BasicBidiStreamingClient/method_empty.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+require_once __DIR__ . '../../../vendor/autoload.php';
+
+// [START bidi_generated_BasicBidiStreaming_MethodEmpty_sync]
+use Google\ApiCore\ApiException;
+use Testing\BasicBidiStreaming\BasicBidiStreamingClient;
+
+
+/**  */
```
bshaffer, line 30: nit - shouldn't have empty block comments.

tests/Unit/ProtoTests/BasicServerStreaming/out/samples/BasicServerStreamingClient/method_empty.php
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * GENERATED CODE WARNING
+ * This file was automatically generated - do not edit!
+ */
+
+require_once __DIR__ . '../../../vendor/autoload.php';
```
bshaffer, line 23: `__DIR__` never has a trailing slash, so the string will need to start with one:
```
require_once __DIR__ . '/../../../vendor/autoload.php';
```

tests/Unit/ProtoTests/Kms/out/samples/V1/KeyManagementServiceClient/get_crypto_key.php
```diff
@@ -0,0 +1,64 @@
...
+
+require_once __DIR__ . '../../../vendor/autoload.php';
+
+// [START cloudkms_v1_generated_KeyManagementService_GetCryptoKey_sync]
+use Google\ApiCore\ApiException;
+use Google\Cloud\Kms\V1\CryptoKey;
+use Google\Cloud\Kms\V1\KeyManagementServiceClient;
+
+/**
+ * Returns metadata for a given [CryptoKey][google.cloud.kms.v1.CryptoKey], as
```
bshaffer, line 31: This format is common in protos, but for PHP specifically we may be better off using [`@see`](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/see.html) tags, as I believe it's more idiomatic. Also the proto name (in this case `google.cloud.kms.v1.CryptoKey`) is not really relevant.

```
 * Returns metadata for a given {@see CryptoKey}, as
```

tests/Unit/ProtoTests/Kms/out/samples/V1/KeyManagementServiceClient/import_crypto_key_version.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * TODO(developer): Replace sample parameters before running the code.
+ */
+function callSample()
+{
+    $formattedParent = KeyManagementServiceClient::cryptoKeyName('[PROJECT]', '[LOCATION]', '[KEY_RING]', '[CRYPTO_KEY]');
+    $algorithm = CryptoKeyVersionAlgorithm::CRYPTO_KEY_VERSION_ALGORITHM_UNSPECIFIED;
+    $importJob = 'import_job';
+    
+    import_crypto_key_version_sample($formattedParent, $algorithm, $importJob);
```
bshaffer, line 78: This is an interesting sample. I think in cases like this, I'd prefer to move this into the sample itself, and make the above place holders the sample params.
```suggestion
    $project = 'project';
    $location = 'location';
    $keyRing = 'key_ring';
    $cryptoKey = 'crypto_key';
    $algorithm = CryptoKeyVersionAlgorithm::CRYPTO_KEY_VERSION_ALGORITHM_UNSPECIFIED;
    $importJob = 'import_job';
    
    import_crypto_key_version_sample($project, $location, $keyRing, $cryptoKey, $algorithm, $importJob);
```

bshaffer (commented):
Preliminary pass with some minor suggestions

src/Generation/SnippetGenerator.php
```diff
@@ -0,0 +1,428 @@
...
+
+class SnippetGenerator
+{
+    /** @var int The license year. */
+    private int $licenseYear;
+
+    /** @var ServiceDetails The service details. */
+    private ServiceDetails $serviceDetails;
+
+    // TODO: existing idioms make constructor private, consider this
```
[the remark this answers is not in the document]
dwsupplee: I can't cite the author's intent, but that would be my assumption as well.

src/Ast/PhpFile.php
```diff
@@ -92,18 +93,27 @@ public function withGeneratedFromProtoCodeWarning(string $filePath, bool $isGa)
...
+
+        if ($this->class) {
+            $lines[] = "namespace {$this->class->type->getNamespace()};\n";
+            $lines[] = $this->uses->toVector()->map(fn ($x) => "use {$x};")->join();
+        }
+
         return implode("\n", $lines) .
             (count($this->uses) >= 1 ? "\n" : '') .
-            static::toPhp($this->class);
+            static::toPhp($this->class ?: $this->block);
```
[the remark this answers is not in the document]
dwsupplee: Good suggestion, I'll add that!

tests/Unit/ProtoTests/BasicBidiStreaming/out/samples/BasicBidiStreamingClient/method_bidi.php
dwsupplee: 👍 , there are a few whitespace issues that I'm working through - my final catch all will be to run [https://github.com/prettier/plugin-php](https://github.com/prettier/plugin-php) on the output, it does a nice job catching any issues that the php-cs-fixer has been having trouble with.

tests/Unit/ProtoTests/Kms/out/samples/V1/KeyManagementServiceClient/get_crypto_key.php
dwsupplee: 👍  - There is a todo here for something like this when possible, https://github.com/googleapis/gapic-generator-php/pull/483/files#diff-6d2583711a56ed5f6ed9262793b7185f0bc017a49a35580b8056429bb2d491fcR161

tests/Unit/ProtoTests/Kms/out/samples/V1/KeyManagementServiceClient/import_crypto_key_version.php
dwsupplee: I'm glad you thought about this too - at first I tried this out, but there were a couple challenges I ran into that made me revert back to this existing approach.

- In the docblock for the main sample, project/location/keyRing etc. wouldn't have any descriptions we could pull in, whereas if we keep the formatted string as the parameter it allows us to leverage the description from the protos: https://github.com/googleapis/gapic-generator-php/blob/e68b44cea965f678731f7f57e2044acfc01ab25b/tests/Unit/ProtoTests/Kms/service.proto#L742-L746
- Some RPCs (at least, outlined in the library service) have multiple formatting methods, which can lead to a really large method signature if we split out each individual component of the resource. It also means we have to take some liberty to programmatically determine what the variable name is and avoid any collisions. I found this was leading to really long, and not entirely helpful variable names - for example: https://github.com/googleapis/gapic-generator-php/blob/e68b44cea965f678731f7f57e2044acfc01ab25b/tests/Unit/ProtoTests/BasicDiregapic/library_rest.proto#L573-L582

src/Generation/SnippetGenerator.php
```diff
@@ -0,0 +1,441 @@
...
+                '// Iterate over pages of elements',
+                AST::assign(
+                    $responseVar,
+                    AST::call(
+                        $snippetDetails->serviceClientVar,
+                        AST::method($method->methodName)
+                    )($snippetDetails->rpcArguments)
+                ),
+                AST::foreach($responseVar->iteratePages(), $page)(
+                    // TODO: figure out how to get the type of the element being iterated over
```
noahdietz, line 229: If the method is Paginated, it will contain a `Type $resourceType` ([link](https://github.com/googleapis/gapic-generator-php/blob/15e8620f0abe3adbc76517962b12445edf90f1ce/src/Generation/MethodDetails.php#L205)) which is the repeated resource message field being iterated over.

Don't need to resolve this TODO in this PR if you don't want to, could follow on.

src/CodeGenerator.php
```diff
@@ -281,6 +281,16 @@ private static function generateServices(
             $code = Formatter::format($code);
             yield ["src/{$version}Gapic/{$service->gapicClientType->name}.php", $code];
 
+            // snippet gen wip
+            $snippetFiles = SnippetGenerator::generate($licenseYear, $service);
+
+            foreach ($snippetFiles as $snippetFile) {
+                $code = $snippetFile[1]->toCode();
+                $code = Formatter::format($code);
+                $methodName = Helpers::toSnakeCase($snippetFile[0]);
```
noahdietz, line 290: Does it make sense to switch the return type of `generate` to `Map` with key = methodName and value = AST or whatever? 

Magic numbers aren't my favorite thing.

dwsupplee: I agree about the numeric index - though the generate method already returns a map instance. When iterating over the map it returns the following data structure (which I'm open to modifying, and would find beneficial unless I am missing something):

```
counter => [
    key, // as set by Map::set($key, $value)
    value
]

e.g.,

// Map::set('hello', world')

0 => ['hello', 'world']

File: src/Generation/SnippetGenerator.php (hunk 9 above)
Comment (dwsupplee) [reply src/Generation/SnippetGenerator.php:229]: I tried that, and it was returning an array type for me - still investigating.

File: src/Generation/SnippetGenerator.php [hunk 11]
```diff
@@ -0,0 +1,441 @@
...
+    public function generateImpl(): Map
+    {
+        $files = Map::new();
+
+        foreach ($this->serviceDetails->methods as $method) {
+            $regionTag = $this->generateRegionTag($method->name);
+            $snippetDetails = new SnippetDetails($method, $this->serviceDetails);
+
+            // TODO: run prettier-php on the resulting files. it'll help clean up any outstanding formatting issues
+            // the existing formatters aren't equipped to catch
```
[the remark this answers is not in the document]
dwsupplee: I wouldn't claim it is overall better, but for our specific use case it is catching some blind spots. The generated output prior to being run through php-cs-fixer is pretty gnarly, and php-cs-fxier does a decent job of formatting - but is having trouble with addressing some unnecessary whitespace. The primary use case I like prettier-php for is it has an option to format lines based on length, which helps to format calls to things like the resource format helpers nicely. I would look at using prettier-php as a stop gap until we can deem whether or not it is worth our time to optimize the raw output from the generator, prior to being run through php-cs-fixer.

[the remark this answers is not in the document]
dwsupplee: I was able to find a line length fixer that integrates with php-cs-fixer, so I went with that.

src/CodeGenerator.php
dwsupplee: I changed this to an array instead of using the Map.

src/Ast/PhpDoc.php
```diff
@@ -415,7 +415,7 @@ protected function toLines(Map $info): Vector
                 $code = Formatter::format("<?php\n{$code}");
                 $code = str_replace('[ELLIPSIS]', '[...]', $code);
                 $code = Vector::new(explode("\n", $code))
-                    ->skip(2)->skipLast(1)
+                    ->skip(1)->skipLast(1)
```
noahdietz, line 418: Nit: could you add a comment for this line that indicates what it is doing? I'm not sure what affect this change has.

src/Ast/PhpFunction.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public function withoutNewlineAfterDeclaration(): PhpFunction
+    {
+        return $this->clone(fn ($clone) => $clone->appendNewline = false);
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
```
noahdietz, line 103: Where is this used?

noahdietz (commented):
Looks great, @dwsupplee

src/CodeGenerator.php
```diff
@@ -281,6 +281,15 @@ private static function generateServices(
             $code = Formatter::format($code);
             yield ["src/{$version}Gapic/{$service->gapicClientType->name}.php", $code];
 
+            // Snippet Generator.
+            $snippetFiles = SnippetGenerator::generate($licenseYear, $service);
+
+            foreach ($snippetFiles as $methodName => $snippetFile) {
+                $code = $snippetFile->toCode();
+                $code = Formatter::format($code, 80);
```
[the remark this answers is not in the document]
bshaffer: To expand on this, looking through the samples there are a few spots where the indentation looks a little aggressive. I'd love to see the samples with line lengths of 100 instead. 

They look good as-is, so if this is something you feel strongly about, I am okay with 80 as well!

src/Ast/PhpFunction.php
dwsupplee: I pulled it over for parity with https://github.com/googleapis/gapic-generator-php/blob/main/src/Ast/PhpMethod.php#L60-L63, I don't think it used right now but could be in the future.

Would you prefer to have it be dropped?

bshaffer (commented):
Some questions from the meeting:

1. Should we have smoke tests for samples?
    - Writing tests is out of scope, but using a static analyzer like PHPStan would be sufficient
3. Where. do the samples end up?
    - They will be at `[ComponentName]/samples/` (e.g. `Vision/samples/`)
5. How do they end up there? 
    - We will upgrade the Bazel build file to the latest version of the generator, and we may need to modify the owlbot repo (or `owlbot.yaml` files) to start copying them over.

src/Ast/PhpBlock.php
```diff
@@ -0,0 +1,48 @@
...
+    public function toCode(): string
+    {
+        $omitSemicolon = false;
+
+        return $this->code
+            ->map(function ($x) use ($omitSemicolon) {
+                $end = $x instanceof PhpBlock
+                    ? ''
+                    : PHP_EOL;
+                return static::toPhp($x, $omitSemicolon) . ($omitSemicolon ? '' : ';') . $end;
```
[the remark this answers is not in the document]
dwsupplee: I agree it is hard to follow and I'm completely open to doing a bit more to think this through and refactor aspects of the AST portion of the codebase, but do you mind if we follow up with that later?

src/Ast/PhpParam.php
```diff
@@ -41,8 +41,15 @@ public function __construct(?ResolvedType $type, Variable $var, $default)
 
     public function toCode(): string
     {
-        $type = is_null($this->type) ? '' : static::toPhp($this->type) . ' ';
-        $default = is_null($this->default) ? '' : ' = ' . static::toPhp($this->default);
+        $type = null;
+        if ($this->type) {
+            $type = substr($this->type->type->name, -2) === '[]'
```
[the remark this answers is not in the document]
dwsupplee: I'll cut this out for now, as it is a remnant of something I refactored out anyways. It may end up being useful for the upcoming work to modify our method signatures for V2, so i'll implement something there if we need it.

src/Generation/FieldDetails.php
```diff
@@ -94,12 +94,17 @@ class FieldDetails
...
     public bool $useResourceTestValue;
 
     /** @var ?int null if not in a one-of; otherwise the index of the one-of - ie every field in a oneof has the same index. */
     public ?int $oneOfIndex;
 
+    /**
+     * @var Vector *Readonly* Vector of FieldDetails. Contains all required subfields on this field, given the field
+     * is a message. Will be empty otherwise.
+     */
+    public Vector $requiredSubFields;
```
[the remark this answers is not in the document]
dwsupplee: > If so, should we refactor the generator to use this field? 

I think it'd be best to touch the gapic generator code when we go to do the work for updating the signatures.

> Will we want to update this to use google.api.method_annotation in the future, instead?

No, I don't think so. This is aso being used in this PR to help populate sample values for required fields of type `message` that may be nested in nature (i.e., not on the top level of the request).

> Also, does this handle OneOfs?

Yes, if the required annotation is applied.

> How does it handle when a field of a nested object is required?

If the subfield is a message, it will recursively process this same logic.

src/Utils/Formatter.php
```diff
@@ -30,74 +36,84 @@ class Formatter
...
+            new Fixer\Import\OrderedImportsFixer(), // -30
+            new Fixer\Whitespace\ArrayIndentationFixer(), // -31
+            new Fixer\Whitespace\MethodChainingIndentationFixer(), // -34
+            new Fixer\Whitespace\SingleBlankLineAtEofFixer(), // -50, PSR2 (must run last)
         ];
 
         // Fixer temporarily removed:
-        // new \PhpCsFixer\Fixer\Whitespace\BlankLineBeforeStatementFixer(), // -21
+        // new Fixer\Whitespace\BlankLineBeforeStatementFixer(), // -21
         // TODO: Understand why this fixer causes too many blank line insertions in some cases.
```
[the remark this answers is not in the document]
dwsupplee: With the changes I added, we may not even need this fixer anymore. Will re-asses in follow up PRs.

src/Utils/Formatter.php
```diff
@@ -145,4 +161,51 @@ public static function moveUseTo(string $codeStr, string $typeName, int $index):
...
+                        new TokenRunner\Analyzer\FixerAnalyzer\IndentDetector(
+                            $whitespacesFixerConfig
+                        ),
+                        $whitespacesFixerConfig
+                    )
+                )
+            ),
+            $blockFinder,
+            new TokenAnalyzer\FunctionCallNameMatcher()
+        );
```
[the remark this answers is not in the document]
dwsupplee: Yes, I did look - but I didn't want to spend too much time reverse engineering their stack to get this fixer working. I will leave a TODO to re-evaluate.

src/CodeGenerator.php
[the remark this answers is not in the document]
dwsupplee: 100 works for me.

tests/Integration/goldens/asset/samples/V1/AssetServiceClient/update_feed.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    // Create a client.
+    $assetServiceClient = new AssetServiceClient();
+
+    // Prepare any non-scalar elements to be passed along with the request.
+    $feedFeedOutputConfig = new FeedOutputConfig();
+    $feed = (new Feed())
+        ->setName($feedName)
+        ->setFeedOutputConfig($feedFeedOutputConfig);
+    $updateMask = new FieldMask();
```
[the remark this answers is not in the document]
dwsupplee: Both of these fields are required on `Feed`, so I don't think one would apply them in the FieldMask. An argument could be made for the oneof value inside of `FeedOutputConfig`, but unfortunately there isn't a great way right now to know when a oneof is required by the server unless they utilize the required annotation (which seems rare).

See the node example: https://github.com/googleapis/nodejs-asset/blob/main/samples/generated/v1/asset_service.update_feed.js#L31-L44

There is some light at the end of the tunnel, though, that should be addressed by phase 2 of snippet gen. We will have configuration files the snippets can read that will apply sensible default values and more concretely handle optional oneofs.

tests/Unit/ProtoTests/Kms/out/samples/V1/KeyManagementServiceClient/import_crypto_key_version.php
bshaffer: We addressed this in the meeting and decided the approach here is preferred, as long as we add a `@see` tag with the resource name helper to the sample PHPDoc.

src/Ast/PhpBlock.php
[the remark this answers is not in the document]
bshaffer: For sure! It's a suggestion but not a particularly important one

src/Ast/PhpDoc.php
dwsupplee: Good call, done!

src/Generation/SnippetGenerator.php
```diff
@@ -0,0 +1,498 @@
...
+                ? $call
+                : AST::assign($var, $call)
+        );
+    }
+
+    /**
+     * @param string $str A string literal to be proxied to AST::literal.
+     * @return AST
+     */
+    private function buildPrintFCall(string $str): AST
```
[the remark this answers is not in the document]
dwsupplee: Nice idea, done.

tests/Integration/goldens/kms/samples/V1/KeyManagementServiceClient/asymmetric_decrypt.php
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Decrypts data that was encrypted with a public key retrieved from
+ * [GetPublicKey][google.cloud.kms.v1.KeyManagementService.GetPublicKey]
+ * corresponding to a [CryptoKeyVersion][google.cloud.kms.v1.CryptoKeyVersion]
+ * with [CryptoKey.purpose][google.cloud.kms.v1.CryptoKey.purpose]
+ * ASYMMETRIC_DECRYPT.
+ *
+ * @param string $formattedName The resource name of the
+ *                              [CryptoKeyVersion][google.cloud.kms.v1.CryptoKeyVersion] to use for
+ *                              decryption.
```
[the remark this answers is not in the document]
dwsupplee: Yes, as we discussed yesterday with the team, that sounds good. Done!

src/Generation/FieldDetails.php
```diff
@@ -260,9 +275,22 @@ public function getOneofDesc()
...
 
-    public function exampleValue(SourceFileContext $ctx)
-    {
-        if ($this->desc->desc->isRepeated()) {
+    /**
+     * Create an example value based off the type of this field.
+     *
+     * @param SourceFileContext $ctx The context this field is attached to.
+     * @param bool $formatStringWithBrackets Whether or not to format a string
+     *             value as "['VALUE']" or "'value'".
```
bshaffer, line 283: nit, but I believe this should be
```suggestion
     *             value as "'[VALUE]'" or "'value'".
```

src/Generation/FieldDetails.php
```diff
@@ -260,9 +275,22 @@ public function getOneofDesc()
...
+     * @param bool $ignoredRepeated If the field is repeated, setting this value
+     *             to true will return a value of the singular type (treats it as non-repeated).
```
bshaffer, line 285: But won't wouldn't this break the sample? `GPBUtil` throws an exception if the provided value isn't an array. We could do something instead like `$wrapRepeated = true` and then wrap the values in an array, e.g.:

```php
$updateMask = [new FieldMask()];
```

Otherwise, static analysis will fail, and samples will be more confusing.

dwsupplee: This was cleared up offline, should be good.

bshaffer: For anyone following along, the reason is that the sample wraps the parameters in an array. This is so sample function arguments can be scalars

bshaffer (approved):

tests/Integration/goldens/dataproc/samples/V1/AutoscalingPolicyServiceClient/create_autoscaling_policy.php
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Creates new autoscaling policy.
+ *
+ * @param string $formattedParent                               The "resource name" of the region or location, as described in
+ *                                                              https://cloud.google.com/apis/design/resource_names. * For
+ *                                                              `projects.regions.autoscalingPolicies.create`, the resource name of the region
+ *                                                              has the following format: `projects/{project_id}/regions/{region}` * For
+ *                                                              `projects.locations.autoscalingPolicies.create`, the resource name of the
+ *                                                              location has the following format: `projects/{project_id}/locations/{location}`
+ *                                                              Please see {@see AutoscalingPolicyServiceClient::regionName()} for help
```
bshaffer, line 43: 😍

tests/Integration/goldens/dataproc/samples/V1/AutoscalingPolicyServiceClient/create_autoscaling_policy.php
```diff
@@ -0,0 +1,124 @@
...
+use Google\Cloud\Dataproc\V1\BasicAutoscalingAlgorithm;
+use Google\Cloud\Dataproc\V1\BasicYarnAutoscalingConfig;
+use Google\Cloud\Dataproc\V1\InstanceGroupAutoscalingPolicyConfig;
+use Google\Protobuf\Duration;
+
+/**
+ * Creates new autoscaling policy.
+ *
+ * @param string $formattedParent                               The "resource name" of the region or location, as described in
+ *                                                              https://cloud.google.com/apis/design/resource_names. * For
```
noahdietz, line 38: Is the spacing here correct? Looks like there is a `* For` which might need to be on a new line?

src/Ast/PhpFunction.php
noahdietz: No that's ok, let's keep it.

tests/Integration/goldens/dataproc/samples/V1/AutoscalingPolicyServiceClient/create_autoscaling_policy.php
bshaffer: Good catch. It seems like these should be bullets [see the autoscaling_policy.proto](https://github.com/googleapis/googleapis/blob/master/google/cloud/dataproc/v1/autoscaling_policies.proto#L261-L270)

amanda-tarafa (approved):
LGTM from the point of view of SnippetGen Phase 1 output.

Thanks!!

tests/Integration/goldens/iam/samples/V1/IAMPolicyClient/get_iam_policy.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+/*
+ * GENERATED CODE WARNING
+ * This file was automatically generated - do not edit!
+ */
+
+require_once __DIR__ . '/../../../vendor/autoload.php';
+
+// [START iam-meta-api_v1_generated_IAMPolicy_GetIamPolicy_sync]
```
amanda-tarafa, line 25: This should be iam only, see b/247779318

```suggestion
// [START iam_v1_generated_IAMPolicy_GetIamPolicy_sync]
```

Also take care to remove the region from a regionalized host if any, [here](https://github.com/googleapis/gapic-generator-csharp/blob/7737012d6f17708c77326059cb9cb0a282e33e59/Google.Api.Generator/Generation/ServiceDetails.cs#L69-L72) the .NET code.

tests/Integration/goldens/iam/samples/V1/IAMPolicyClient/get_iam_policy.php
```diff
@@ -0,0 +1,64 @@
...
+use Google\ApiCore\ApiException;
```
amanda-tarafa, line 26: Also, at some point inside the region tag, there should be a notice as describe in b/240118241. This notice should appear on all SnippetGen Phase 1 samples.

amanda-tarafa (changes requested):
Undoing LGTM, as I just thought to check for two issues that were recently surfaced on all other generators. It's worth it to ship PHP with them fixed.

Thanks!

noahdietz (approved):
General LGTM

tests/Integration/goldens/dataproc/samples/V1/AutoscalingPolicyServiceClient/create_autoscaling_policy.php
dwsupplee: Good eye Noah. I think I know exactly what caused this - running the docblock through `PhpDoc::text` to help format the line with the addition of the message to take a look at the formatting method. Taking a look at options now.

tests/Integration/goldens/asset/samples/V1/AssetServiceClient/analyze_iam_policy.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * TODO(developer): Replace sample parameters before running the code.
+ */
+function callSample(): void
+{
+    $analysisQueryScope = '[SCOPE]';
+
+    analyze_iam_policy_sample($analysisQueryScope);
+}
+// [END cloudasset_v1_generated_AssetService_AnalyzeIamPolicy_sync]
```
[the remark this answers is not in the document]
dwsupplee: Good call out 👍. The guideline does include the helper method, for reference: https://googlecloudplatform.github.io/samples-style-guide/#easy-run-function

tests/Integration/goldens/iam/samples/V1/IAMPolicyClient/get_iam_policy.php
dwsupplee: Thank you, this is all set! :)

tests/Integration/goldens/dataproc/samples/V1/AutoscalingPolicyServiceClient/create_autoscaling_policy.php
dwsupplee: Should be resolved now 👍

amanda-tarafa (approved):
LGTM. Thanks!

PR #485, opened 2022-06-14
chore: rename client to gapicClient in generated tests

dwsupplee:
This rename helps us to avoid conflicts in the generated tests when a proto field uses the name `client`.

The second commit here removes the asset_rest goldens, as it looks like the supporting code to update them was [removed from the toolchain and replaced with the compute_small set of tests](https://github.com/googleapis/gapic-generator-php/pull/278). If you'd prefer we don't remove this here, I can drop out the commit.


src/Generation/UnitTestsGenerator.php
```diff
@@ -298,7 +298,7 @@ private function testExceptionalCaseNormal(MethodDetails $method): PhpMethod
                 $requestPerField->map(fn ($x) => $x->initCode),
                 AST::try(
                     $client->instanceCall(AST::method($method->methodName))($requestCallArgs),
-                    '// If the $client method call did not throw, fail the test',
+                    '// If the $gapicClient method call did not throw, fail the test',
```
noahdietz, line 301: Nit: This is kind of a hack, but you could put `$client->toCode()` here and it would emit `$gapicClient`, so that we don't have a hard coded value? Or perhaps even try just putting the `$client` var in there and see if it generates it properly (without you having to call `toCode()` directly). Alternatively, just use a variable for the AST::var name and for here.

Kind of applies elsewhere. lmk wyt

noahdietz (commented):
One nit, but looks good to me. Removing that asset_rest baseline is fine, thank you for cleaning that up.

bshaffer (approved):

src/Generation/UnitTestsGenerator.php
dwsupplee: I'll update to use a variable, good call out.

dwsupplee: We could probably do some very similar changes for other fields to help improve maintainability, let's reserve that for future updates.

noahdietz: Yeah good point. Thank you!

noahdietz (approved):

PR #493, opened 2022-07-19
feat: enable generation of LongRunning GAPIC

bshaffer:


src/CodeGenerator.php
```diff
@@ -155,7 +155,12 @@ public static function generate(
                     $serviceDetails =
                         new ServiceDetails($catalog, $namespaces[0], $fileDesc->getPackage(), $service, $fileDesc, $transportType);
                     $serviceName = $serviceDetails->serviceName;
-                    if (!in_array($serviceName, self::MIXIN_SERVICES)) {
+                    $generateNormalGapic = !in_array($serviceName, self::MIXIN_SERVICES);
+                    if (!$generateNormalGapic) {
+                        $generateNormalGapic = $serviceYamlConfig->apiNames->contains($serviceName)
```
noahdietz, line 160: Can you add a comment here to the effect of "if we want to generate a gapic for a mixin service, this is how we differentiate that intention and the integration of a mixin service".

noahdietz (approved):

PR #494, opened 2022-07-20
fix(bazel): support changing the generator

noahdietz:
Support supplying a different `gapic-generator-php` binary to the `php_gapic_library` target. This is a mirror of what [C# did](https://github.com/googleapis/gapic-generator-csharp/pull/480) to support their major version rollout.

For http://b/239612041.


vam-google (approved):
LGTM

PR #510, opened 2022-08-17
feat: switch to startApiCall with call descriptors

noahdietz:
_This PR is made against the `v2-dev` branch from which we can merge into `main` as needed._

Add generation of the following descriptor config fields:
* `callType`: populated with `\Google\ApiCore\Call::{RPC}_TYPE` constant
* `responseType`: populated with the response message PHP type name
* `headerParams`: (if configured) populated with the header request params (implicit or explicit) config
* `interfaceOverride`: (if mix-in) the name of the proto service that owns the mix-in RPC

These descriptor config fields are only used by `GapicClientTrait->startApiCall`.

Switches generation of all method types to invoke `startApiCall` instead of the call type-specific gax APIs. This means header injection is removed from the GAPIC layer and pushed into the GAX layer (already supported).


src/Utils/ProtoHelpers.php
```diff
@@ -105,6 +109,84 @@ public static function routingParameters(ProtoCatalog $catalog, ?DescriptorProto
...
+                    'keyName' => $key
+                ]
+            )->values()->toArray();
+
+            $headerParams = array_merge($headerParams, $flattened);
+        }
+
+        return $headerParams;
+    }
+
```
noahdietz, line 189: Most of this will be refactored when we rip out the header param injection code while moving to `startApiCall`. A deep review is probably not necessary.

noahdietz:
That was nasty, there was a leading whitespace in the title that made conventionalcommit bot unhappy

noahdietz:
This should be ready for review now with the added `interfaceOverride` changes.

src/Generation/GapicClientGenerator.php
```diff
@@ -825,79 +783,32 @@ private function rpcMethod(MethodDetails $method): PhpClassMember
...
+        // Fall through to CLIENT_STREAMING.
       case MethodDetails::CLIENT_STREAMING:
-        $startCallArgs[] = AST::NULL;
-        $startCallArgs[] =
-          AST::access($this->ctx->type(Type::fromName(Call::class)), AST::constant('CLIENT_STREAMING_CALL'));
-        if ($method->isMixin()) {
-            $startCallArgs[] = $method->mixinServiceFullname;
-        }
-        return AST::call(AST::THIS, AST::method('startCall'))(...$startCallArgs);
+        $startApiCallArgs = $startApiCallArgs->set('request', AST::NULL);
```
dwsupplee, line 796: I realize the indentation was off prior to this change, but would you be open to cleaning it up here?

tests/Unit/Utils/ProtoHelpersTest.php
```diff
@@ -18,6 +18,7 @@
 
 namespace Google\Generator\Tests\Unit\Utils;
 
+use Google\Generator\Collections\Map;
```
dwsupplee, line 21: This import looks unused

src/Utils/ProtoHelpers.php
```diff
@@ -78,31 +79,81 @@ public static function isMap(ProtoCatalog $catalog, FieldDescriptor $desc): bool
...
-                fn ($x) => $x->getField(),
-                // Value: The routing header config the RoutingParameter defines.
-                fn ($x) => [
-                    'getter' => static::buildGetterChain($catalog, $msg, $x->getField()),
-                    'key' => static::fieldOrTemplateVariable($x),
-                    'regex' => static::compileRoutingRegex($x),
-                    'root' => explode('.', $x->getField())[0],
-                ],
+
+        // Organzie rules by resulting key name.
```
dwsupplee, line 115:
```suggestion
        // Organize rules by resulting key name.
```

src/Utils/ProtoHelpers.php
```diff
@@ -78,31 +79,81 @@ public static function isMap(ProtoCatalog $catalog, FieldDescriptor $desc): bool
...
+        // Organize "compiled" segment matchers by resulting key name.
+        $matchersByKey = $rulesByKey->mapValues(
+            fn($key, $routingRules) =>
+                $routingRules
+                    ->filter(fn($rule) => static::hasMatcher($rule))
+                    ->map(fn($rule) => static::compileRoutingRegex($rule)))
+            ->filter(fn($key, $matchers) => $matchers->count() > 0)
+            ->mapValues(fn($key, $val) => $val->toArray());
+
+        // Reduce the rules for each key into one rule, taking the first it's
```
dwsupplee, line 128:
```suggestion
        // Reduce the rules for each key into one rule, taking first its
```

src/Utils/ProtoHelpers.php
```diff
@@ -78,31 +79,81 @@ public static function isMap(ProtoCatalog $catalog, FieldDescriptor $desc): bool
...
+        return $paramDescs->toArray();
+    }
+
+    private static function hasMatcher(RoutingParameter $routingParam): bool
+    {
+        // Has a path_template in RoutingParameter and the path_template defines
+        // a segment matcher other than '**' ("match anything"), with the header
+        // key override.
+        $temp = $routingParam->getPathTemplate();
+        return !empty($temp) && str_contains($temp, '=') && !str_ends_with($temp, "=**}");
```
dwsupplee, line 156: str_contains / str_ends_with were introduced in PHP 8, if we'd like to keep compatibility with 7.4 we'll need to replace these.

dwsupplee (changes requested):
Looking good, just a few minor notes.

src/Generation/GapicClientGenerator.php
noahdietz: Ha I don't mind at all, I was wondering why it was like that assuming the formatter did it. Done.

tests/Unit/Utils/ProtoHelpersTest.php
noahdietz: Indeed it is! Removed.

src/Utils/ProtoHelpers.php
bshaffer:
```suggestion
        return !empty($temp) && false !== strpos($temp, '=') && substr($temp, -4) !== '=**}';
```

noahdietz: Ah good point. I've added some implementations with TODOs to remove them once we switch this repo to PHP 8.x. I also changed another spot that used these PHP8 methods.

noahdietz: Ha @bshaffer we replied almost at the same time. I just added some implementations for these methods and replaced them so that read the same. WDYT?

bshaffer: Your solution looks way better ;) nice work!

dwsupplee (approved):

bshaffer (approved):

PR #522, opened 2022-10-06
chore: v2 dev catch up to main

noahdietz:
Catches up `v2-dev` with changes from `main` so that we can merge `v2-dev` into `main` for the `startApiCall` refactoring.

I had to updated the golden files for both unit and integration tests.


noahdietz:
I'm not 100% sure this is the right way to update `v2-dev`, but there were many conflicts with `main` and then I had to update the golden files so I just opened a PR.

noahdietz:
I think this might actually mess up the history...idk. I might just do an in place merge and baseline update on v2-dev instead

noahdietz:
Yep that worked much better: https://github.com/googleapis/gapic-generator-php/pull/524

PR #524, opened 2022-10-07
feat: generate call descriptors and use startApiCall

noahdietz:
This refactors the generated client implementation to use the `startApiCall` invoker from `gax`. This also includes generation of extra call descriptors in the config files that described the call type and all properties.

This is an entirely internal refactor and does not affect the public surface.


noahdietz:
Verified with some sample tests that things still work with these changes :D

.github/workflows/bazel_example.yml
```diff
@@ -2,9 +2,9 @@ name: Bazel Build
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, v2-dev ]
```
bshaffer, line 5: We'd want to remove this before merging into main, yes?

bshaffer: Now I'm thinking we want to keep it, since we're keeping the `v2-dev` branch even after this is merged.

noahdietz: yeah I think we should keep it if we are keeping the branch.

noahdietz:
Looks like I need to update the bazel goldens after catching up with `main`

noahdietz:
Closing this PR as we've decided to move these changes to the separate v2 client. @dwsupplee will be migrating the work.

PR #531, opened 2022-10-31
chore: rename serviceAddress to apiEndpoint

bshaffer:
We have code in [each owlbot.py](https://github.com/googleapis/google-cloud-php/blob/main/AnalyticsAdmin/owlbot.py#L33-L49) to rename `$serviceAddress` to `$apiEndpoint`, to mark `$serviceAddress` as deprecated, and to rename an instance of `$serviceAddress` in the parameter description. 

This happens in every GAPIC. I verified, and there are none that do not do this.

I suggest we remove the parameter all together. This will have the following effects:

 - We can simplify our `owlbot.py` for every package by removing the above block above.
 - In all cases, the [second replacement](https://github.com/googleapis/google-cloud-php/blob/main/AccessContextManager/owlbot.py#L45-L48) of `$serviceAddress` in `owlbot.py` failed (see [AccessContextManagerGapicClient](https://github.com/googleapis/google-cloud-php/blob/main/AccessContextManager/src/V1/Gapic/AccessContextManagerGapicClient.php#L486)) because of a linebreak changing the expected regex.
 - We can safely make this change without worrying about compatibility, since this removes only the documentation of the parameter, and not the logic handing the parameter itself.

TODO: Remove the `serviceAddress` parameter from GAX as well


noahdietz (approved):
Will we need to fix every owlbot config to remove the find-replace in conjunction with this change?

bshaffer:
> Will we need to fix every owlbot config to remove the find-replace in conjunction with this change?

No, because this change essentially does what the owlbot config did (with the exception that we will no longer document the deprecated parameter). But we still should go and remove all the owlbot config code from the other components once this rolls out

PR #540, opened 2022-12-01
fix: miscellaneous snippet gen fixes

dwsupplee:
While going through the influx of new PRs generated by snippet gen I discovered a handful of issues which this PR addresses.

Each commit address a separate issue, if preferred I can open multiple PRs - but I thought it would be nice to have the test updates all in one place.

Here are the learnings/fixes:

1. Format methods aren't only applied at the base request level, they can be found in nested messages as well. Adjusted for this.
2. Updated inline var block on `getResults`calls in LRO samples to annotate the correct item.
3. Prevent calling `serializeToJsonString` on primitives.

I also temporarily disabled the inline var block for paginated samples on REST only clients. When the resource type is a map the `MethodDetails`class is having trouble determining the correct type to display. Seems like a wider issue that I plan to address separately.

### DIFF
--- a/src/Generation/SnippetGenerator.php
+++ b/src/Generation/SnippetGenerator.php
@@ -175,7 +175,7 @@ class SnippetGenerator
                             : Vector::new([
                                 AST::inlineVarDoc(
                                     $context->type($snippetDetails->methodDetails->lroResponseType),
-                                    $responseVar
+                                    $resultVar
                                 ),
                                 AST::assign($resultVar, $responseVar->getResult()),
                                 $this->buildPrintFCall(
--- a/src/Generation/SnippetGenerator.php
+++ b/src/Generation/SnippetGenerator.php
@@ -160,6 +160,7 @@ class SnippetGenerator
         $errorVar = AST::var('error');
         $isCustomOp = $snippetDetails->methodDetails->methodType === MethodDetails::CUSTOM_OP;
         $context = $snippetDetails->context;
+        $lroResponseType = $snippetDetails->methodDetails->lroResponseType;
 
         return $this->buildSnippetFunctions(
             $snippetDetails,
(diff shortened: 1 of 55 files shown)


src/Main.php
```diff
@@ -150,6 +150,7 @@ function showUsageAndExit()
         $gapicYaml,
         $serviceYaml,
         $numericEnums,
+        -1,
```
noahdietz, line 153: Sorry, what is this for?

src/Generation/SnippetGenerator.php
```diff
@@ -244,13 +252,15 @@ private function rpcMethodExampleBidiStreaming(SnippetDetails $snippetDetails):
...
+                    $context->type($responseType),
                     $elementVar
                 ),
                 AST::foreach($streamVar->closeWriteAndReadAll(), $elementVar)(
                     $this->buildPrintFCall(
                         'Element data: %s',
-                        "{$elementVar->toCode()}->serializeToJsonString()"
+                        $responseType->isClass()
+                            ? "{$elementVar->toCode()}->serializeToJsonString()"
+                            : $elementVar->toCode()
```
noahdietz, line 263: Curious, can you share the example that necessitates this code? I think all LRO response types must be a `Message`, so I'm wondering when we would ever _not_ use `serializeToJsonString`.

bshaffer (approved):
This looks great! One small request

src/Generation/SnippetGenerator.php
dwsupplee: I think you're right, I put this here for safety just in case. If you prefer, I'm happy to remove it 👍 .

src/Generation/SnippetGenerator.php
```diff
@@ -207,20 +205,29 @@ private function rpcMethodExamplePaginated(SnippetDetails $snippetDetails): AST
...
                 $this->buildClientMethodCall($snippetDetails, $responseVar),
                 PHP_EOL,
-                AST::inlineVarDoc(
-                    $context->type($snippetDetails->methodDetails->resourceType),
-                    $elementVar
-                ),
+                // When transport is REST only, disabling this for now.
+                // Need to further investigate an issue causing the resourceType
+                // to render as ItemsEntry with a mapped entry, despite a
+                // different value being outlined in the proto.
```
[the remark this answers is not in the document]
dwsupplee: Yes! https://github.com/googleapis/gapic-generator-php/issues/544

src/Generation/SnippetGenerator.php
noahdietz: Hm...I'm kind of in favor of removing it so as to reduce complexity since the case it handles is not actually possible. A `google.protobuf.Any` can only contain a `message`.

src/Main.php
noahdietz: Oh ok...maybe this is a "nit" but could you add `/* $licenseYear */` or something to this so that we know what it is? Probably not a big deal, just looked out of place in the context of this CL.

src/Generation/SnippetGenerator.php
dwsupplee: All set on LRO! This comment is on the bidi sample, so it won't auto resolve.

src/Main.php
dwsupplee: Sure thing

noahdietz (approved):

PR #543, opened 2022-12-05
feat: introduce v2 client with new method signatures

dwsupplee:
This PR includes the changes in https://github.com/googleapis/gapic-generator-php/pull/542. Once that is merged I'll update here and the diff should clean up a little bit.

I opted to dupe a bit of the code between the unit test/gapic generators into the "V2" based classes in lieu of engineering out something more DRY for the following reasons:

- We aren't 100% on whether the new client will be the final approach
- If this is indeed final approach, we will likely remove the "legacy" generators in short order anyways

Open to discussion on this if anyone feels like that wasn't the right call 👍 .

I'll update the goldens for the integration tests once https://github.com/googleapis/gapic-generator-php/pull/542 / https://github.com/googleapis/gapic-generator-php/pull/540 are merged.

Please note, don't take the way the new client looks as final (e.g., the base client is still a class and not abstract) - still plenty of room for discussion on how we want that to look. The primary purpose of this PR is to move forward with how the new signatures will look/work.


dwsupplee:
The remaining failing tests are due to lack of formatting methods in the v2 gapic, which we will be tagging on with some new work @noahdietz is doing.

noahdietz (commented):
loving it so far!

noahdietz (approved):
:shipit:

PR #548, opened 2022-12-14
fix(snippetgen): fix field name typo

noahdietz:
Fix typo in field name reference causing http://b/262576070.

The branch wasn't being covered by tests, and to ensure that it was, I had to rename a proto service from `LibraryService` to `Library`, then add a wonky comment to a field with a resource_reference to ensure the code path was exercised.

We need to break logic out into unit tests to make this easier.


dwsupplee (approved):
Tricky solution with the rename of the service, great idea!

Issue #453, opened 2022-02-08
Generate resource name helpers for resources referenced at any level of request messages


PierrickVoulet:
**Is your feature request related to a problem? Please describe.**

The generator generates resource name helpers for resources that are referenced in top-level messages _only_.

The Google Ads API contains resources that are only referenced deep down in request messages. The library `google-ads-php` for the Google Ads API relies heavily on the generated resource name helpers.

**Describe the solution you'd like**

Generate resource name helpers for resources referenced at any level of request messages, not only in top-level messages.

**Describe alternatives you've considered**

Historically, we worked around this limitation in the library `google-ads-php` by manually implementing resource name helpers that were not automatically generated but we now have too many of those (60+) to continue working with this workaround.

**Additional context**

The issue was initially reported at b/161259389, you can find more context and some examples there.

PierrickVoulet:
There is another consequence for this limitation:

Let's say I have a top-level message X that triggers the generation of resource name helper Y and my library users leverage this last to build or parse resource names. This is typically useful in unit tests or to build unrelated requests that reference the same type of resource name at lower levels. Editing the message X in future versions can lead to the resource name helper Y to disappear, this can lead to unnecessary breaking changes for the users who rely on it.

noahdietz:
I'm not sure what kinds of changes would result in a change to the generation of the helper and not also be a plain ol'breaking change in the proto surface. The situations I can think of that would result in a resource name helper no longer being generated are also standard breaking changes to the proto surface (e.g. field type changing, removal of a field, removal of an RPC, etc.) AFAIK. Did you have a specific scenario in mind?

PierrickVoulet:
Correct, changes that lead resource name helpers to not be generated anymore would be part of a major release.

My point is that this FR would prevent breaking the users of those generated resource name helpers who use them for messages that are not at top-level.

PierrickVoulet:
Please see https://github.com/googleads/google-ads-php/issues/747#issuecomment-1046058571 for a real life example of the type of unnecessary breaking changes that I was referring to in my previous comment.

fiboknacky:
@noahdietz Noah, any chances this could be prioritized this year?

noahdietz:
@fiboknacky yeah, this is actually done by #551:
> Increases the request message search depth for resource references/definitions.

[...]

PR #551, opened 2022-12-15
feat(v2): generate ResourceHelperTrait and helpers

noahdietz:
Introduces generation of the `ResourceHelperTrait` & v1 backwards compatible formatters in **v2** clients for those services that make use of resource names in their request surface. This also generates the `templateMap` into the client resource descriptor to be loaded by the `ResourceHelperTrait`.

Increases the request message search depth for resource references/definitions.

This also increases the version of gax-php to 1.18 in order to leverage the `ResourceHelperTrait`.

Updates unit & integration test goldens.

Refactors direct use of `ProtoHelpers::getCustomOption` to annotation-specific helpers.


tests/Integration/goldens/dataproc/src/V1/Client/BaseClient/AutoscalingPolicyServiceBaseClient.php
```diff
@@ -91,6 +93,130 @@ private static function getClientDefaults()
...
+    {
+        return self::getPathTemplate('region')->render([
+            'project' => $project,
+            'region' => $region,
+        ]);
+    }
+
+    private static function registerPathTemplates()
+    {
+        self::loadPathTemplates(__DIR__ . '/../../resources/autoscaling_policy_service_descriptor_config.php', self::SERVICE_NAME);
```
bshaffer, line 189: nit but can we break this up?
```suggestion
        self::loadPathTemplates(
            __DIR__ . '/../../resources/autoscaling_policy_service_descriptor_config.php',
            self::SERVICE_NAME
        );
```

We should just need to take advantage of the [Formatter:: buildLineLengthFixer](https://github.com/googleapis/gapic-generator-php/blob/main/src/Utils/Formatter.php#L166) method by passing in `$lineLength`.

tests/Integration/goldens/dataproc/src/V1/Gapic/ClusterControllerGapicClient.php
```diff
@@ -136,6 +148,116 @@ private static function getClientDefaults()
         ];
     }
 
+    private static function getClusterNameTemplate()
```
bshaffer, line 151: Just for my own understanding, why weren't these showing up before? I would have expected to see a lot of deletions from the old properties and methods being removed.

noahdietz: This is likely being added because this PR expands the resources considered "in scope" by traversing the entire request message structure for any message or field representing a resource name to ensure that it has a helper generated. See #453

tests/Integration/goldens/dataproc/src/V1/Client/BaseClient/AutoscalingPolicyServiceBaseClient.php
noahdietz: Oops i shouldn't have committed that haha I will do as you suggested but in the generator code.

noahdietz: What kind of `$lineLength` do you want? I will have to use it for the entire v2 GAPIC client file.

noahdietz: To get this to reliably wrap, we probably need to target `80`

dwsupplee: When adding the V2 clients I tried to add a line length formatter, but it did some odd formatting I wasn't super happy with. I'd like to do a follow up PR to address, so I wouldn't worry too much about the formatting right now.

noahdietz: Thanks I think that makes sense...trying `80` did make some weird things...

noahdietz: We should look at doing more localized formatting in the `AST::block` code gen  maybe, for things like this

tests/Integration/goldens/dataproc/src/V1/Gapic/ClusterControllerGapicClient.php
bshaffer: That makes sense, but wouldn't we still have _some_ coverage for the resource name helpers which were already being generated?

bshaffer: For instance, why aren't these methods being removed? 

https://github.com/googleapis/gapic-generator-php/blob/v2-dev/tests/Integration/goldens/asset/src/V1/Gapic/AssetServiceGapicClient.php#L143

noahdietz: Why would they be removed? We aren't taking a breaking change for these until we do the major version rev

bshaffer: They're private methods, so they can be removed. I thought we were moving the previous method of generating resource name templates over to using the descriptors. 

So we're only moving to the `ReosurceHelperTrait` for:
 1. V2 clients
 2. V1 clients for previously out-of-scope resource name helpers
Is that correct?

noahdietz: I thought we weren't changing V1 Clients at all...literally leaving them the exact same, untouched, no refactoring whatsoever ?

bshaffer: OK, I understand. The reason we are doing this is because it requires a GAX version that is incompatible with PHP 5.6.

It's a point to be discussed later, but I do think we should refactor V1 clients to use both `ResourceHelperTrait` and `startApiCall` before we drop support for V1. Otherwise we'll have to maintain both types of calls (using `ResourceNameTrait` vs legacy method, and using `startApiCall` vs legacy method) for the entire compatibility period of about 6 months.

bshaffer: This would have to be after we drop support for PHP 5.6

dwsupplee (approved):
Looks great, thanks Noah!

PR #554, opened 2022-12-20
feat(v2): add sendAsync generation support

noahdietz:
Adds generation of the GAPIC v2 method `sendAsync`. This method invokes the `GapicClientTrait.startAsyncCall` to start the API call, then returning the `Promise` directly for the caller to manage. This only works for non-streaming calls. This method will only be generated for GAPICs with at least one non-streaming RPC. 

Also add to convenience methods to `ResolvedType` for the PHP types `string` and `array`.

### DIFF
--- a/src/Generation/GapicClientExamplesGenerator.php
+++ b/src/Generation/GapicClientExamplesGenerator.php
@@ -133,6 +133,34 @@ class GapicClientExamplesGenerator
         ];
     }
 
+    private static function initRequestMessage(MethodDetails $method, SourceFileContext $ctx): AST
+    {
+        // TODO: Switch this to request builder when possible.
+        return AST::new($ctx->type($method->requestType))();
+    }
+
+    public static function sendAsyncExample(ServiceDetails $service, SourceFileContext $ctx): AST
+    {
+        $serviceClient = AST::var($service->clientVarName);
+        $method = $service->methods->firstOrNull(fn($m) => !$m->isStreaming());
+        if (!$method) {
+            return null;
+        }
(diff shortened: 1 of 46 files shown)


bshaffer (approved):
This looks great! Somewhat related: 

We could release a [v1.15.1](https://github.com/googleapis/gax-php/releases) version of GAX that supports PHP 5.6, which implements the methods `GapicClientTrait::startAsyncCall` and which just throws an exception like `to use this method, you must be running on PHP 7.0 and above, and using GAX v1.18 or higher`. We would do the same thing in `loadPathTemplates`

noahdietz:
> version of GAX that supports PHP 5.6, which implements the methods `GapicClientTrait::startAsyncCall`

Sorry, why would we want/need to do this?

bshaffer:
> Sorry, why would we want/need to do this?

This would just allow us to release the new (v2) clients while keeping PHP 5.6 support in google-cloud-php, and giving a better experience to users on PHP 5.6 by giving them a PHP fatal error.

We would then update `composer.json` in google-cloud-php to be something like `^1.15.1|^1.18.1`.

noahdietz:
> This would just allow us to release the new (v2) clients while keeping PHP 5.6 support in google-cloud-php, and giving a better experience to users on PHP 5.6 by giving them a PHP fatal error.

Aw, yes that's a good idea. As we discussed offline, your suggestion is definitely the easiest "process-wise". Let's discuss it some more, and not block this PR on this point. Just need sign off from @dwsupplee.

noahdietz:
Decided to go with #562

PR #571, opened 2023-02-07
feat: add request builder snippets

bshaffer:


tests/Unit/ProtoTests/BasicDiregapic/library_rest.proto
```diff
@@ -187,7 +187,7 @@ service Library {
     };
     option (google.api.method_signature) = "name,book";
     option (google.api.method_signature) =
-        "name,optional_foo,book,update_mask,physical_mask";
```
bshaffer, line 190: removed `physical_mask` because this field does not exist in the protos, and so was throwing an exception. I believe this is the behavior we want.

src/CodeGenerator.php
```diff
@@ -367,6 +368,15 @@ private static function generateServices(
...
             yield ["src/{$version}resources/{$service->clientConfigFilename}", $json];
             // TODO: Further files, as required.
+            // Resource: build_method.txt
+            $ctx = new SourceFileContext($service->gapicClientType->getNamespace(), $licenseYear);
+            $buildMethodFragments = BuildMethodFragmentGenerator::generate($ctx, $service);
+            foreach ($buildMethodFragments as [$fragmentName, $buildMethodFragments]) {
+                $buildMethodFragmentCode = BuildMethodFragmentGenerator::format(
+                    $buildMethodFragments->reduce('', fn ($v, $i) => $v . $i->toCode())
+                );
+                yield ["fragments/{$version}{$fragmentName}.build.txt", $buildMethodFragmentCode];
```
noahdietz, line 378:
```suggestion
                yield ["fragments/{$fragmentName}.build.txt", $buildMethodFragmentCode];
```

Paired with the other suggested change in BuildMethodFragmentGenerator

src/Generation/BuildMethodFragmentGenerator.php
```diff
@@ -0,0 +1,214 @@
...
+                    if (empty($methodSignature)) {
+                        continue;
+                    }
+                    $buildMethods = $buildMethods->append(
+                        $this->buildMethodSnippet($methodDetails, $methodSignature, 0 === $i)
+                    );
+                }
+                if ($buildMethods->count() > 0) {
+                    $buildMethodSnippets = $buildMethodSnippets->set(
+                        $methodDetails->requestType->name,
```
noahdietz, line 70:
```suggestion
                        str_replace('\\', '/', $methodDetails->requestType->getFullname(true)),
```

So the fragements are `Google/Cloud/Foo/V1/GetFooRequest.build.txt`

noahdietz:
Can you run the GoldenUpdater now to update all of the unit tests? Your code for fragment generation looks good.

noahdietz:
Looks like there is an edge case (when there is no method_signature) that isn't covered
```
--gapic_out: Error from PHP gapic generator:
missing method signature arguments
```

src/Generation/BuildMethodFragmentGenerator.php
```diff
@@ -0,0 +1,214 @@
...
+            }
+        };
+
+        $methodSignatureArguments = array_filter(explode(',', $methodSignature));
+        $requiredFields = $methodDetails->allFields
+            ->filter(fn ($f) => in_array($f->name, $methodSignatureArguments))
+            ->orderBy(fn ($f) => array_search($f->name, $methodSignatureArguments));
+
+        if (count($methodSignatureArguments) !== $requiredFields->count()) {
+            throw new \LogicException('missing method signature arguments');
```
noahdietz, line 132: Let's make this error more descriptive so its easier to debug e.g. include the method name and signature being evaluated.

bshaffer: done!

tests/Integration/goldens/asset/fragments/Google/Cloud/Asset/V1/SearchAllIamPoliciesRequest.build.txt
```diff
@@ -0,0 +1,106 @@
...
+     *                               word in any of the searchable fields (except for the included
+     *                               permissions).
+     *                               * `resource:(instance1 OR instance2) policy:amy` to find
+     *                               IAM policy bindings that are set on resources "instance1" or
+     *                               "instance2" and also specify user "amy".
+     *                               * `roles:roles/compute.admin` to find IAM policy bindings that specify the
+     *                               Compute Admin role.
+     *                               * `memberTypes:user` to find IAM policy bindings that contain the "user"
+     *                               member type.
+     * @param mixed  $optionalFields {
```
bshaffer, line 56: I am thinking it may make more sense to not include the optional fields in the builders. This would encourage the use of fluid setters as opposed to a big array of properties:

```php
// using fluid setters
$request.= SearchAllIamPoliciesRequest::build('scope', 'query')
    ->setAssetTypes($assetTypes);
// using array of "optionalFields"
$request.= SearchAllIamPoliciesRequest::build('scope', 'query', ['asset_types' => $assetTypes]);

```

noahdietz: That is fine with me, and those fields can still be set after the request is created. The primary use case for these builders/`method_signature` is to name the bare minimum of fields necessary to make a successful request. SGTM.

bshaffer: Done!

phpunit.xml.dist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
noahdietz, line 1: What is this file about anyways?

noahdietz (approved):
@bshaffer random question: Do you know if the fragment files will be copied by OwlBot into the repo PRs if they exist in the `php_gapic_assembly_pkg`-created tarball? Put differently, should we be excluding those fragment files from the generated tarball? If so, we can fix it in a separate PR.

phpunit.xml.dist
bshaffer: It just allows you to run `phpunit` in the root without having to specify the location of the tests. It's best practice to have one! I didn't realize I added one in this PR though, so I can remove it :P

bshaffer:
> should we be excluding those fragment files from the generated tarball? If so, we can fix it in a separate PR.
That sounds like a good idea! That would ensure that if something goes wrong or is misconfigured, the files never end up in the result.

phpunit.xml.dist
noahdietz: eh its fine! Keep it.

noahdietz:
> > should we be excluding those fragment files from the generated tarball? If so, we can fix it in a separate PR.
> 
> That sounds like a good idea! That would ensure that if something goes wrong or is misconfigured, the files never end up in the result.

Cool, I will follow up this PR with another to ensure that the fragments don't end up in the post-processed result.

noahdietz (approved):

PR #584, opened 2023-03-03
fix: dupe method names from mixins

### LINKED ISSUE #561: Ensure mixin RPCs only generate a single method on a generated client
When a mixin RPC (IAM, for example) is defined on both the proto and in a service's yaml configuration, the generator is currently erroneously generating two methods. The mixin RPC defined on the proto should take precedent and be the only method defined.

Examples:
https://github.com/googleapis/googleapis/blob/master/google/cloud/dataplex/v1/content.proto#L94-L103
https://github.com/googleapis/googleapis/blob/master/google/cloud/dataplex/v1/dataplex_v1.yaml#L31-L35

Generates:
https://github.com/googleapis/google-cloud-php/blob/5cacff7458b3e69c4992d4fa410918bed251c648/Dataplex/src/V1/Gapic/ContentServiceGapicClient.php#L510-L532
https://github.com/googleapis/google-cloud-php/blob/5cacff7458b3e69c4992d4fa410918bed251c648/Dataplex/src/V1/Gapic/ContentServiceGapicClient.php#L857-L881

bshaffer:
fixes #561

This is the correct way to handle it according to https://google.aip.dev/client-libraries/4234#overriding-a-duplicate-rpc

### DIFF
--- a/src/Generation/ServiceDetails.php
+++ b/src/Generation/ServiceDetails.php
@@ -174,11 +174,11 @@ class ServiceDetails
             // Technically there could be multiple different named operation services,
             // but for simplicity we will assume they are all the same and use the first.
             $this->customOperationService = $customOperations[0]->operationService;
-            
+
             // Determine if the operation service implements the Cancel and/or the Delete RPCs.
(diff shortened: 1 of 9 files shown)


noahdietz (approved):
Great catch

noahdietz (approved):

PR #585, opened 2023-03-03
fix: max precision for float test values

bshaffer:
Fixes the errors thrown in our tests here: https://github.com/googleapis/google-cloud-php/pull/5704/files#r1102266788

Protobuf PHP does not allow more than 2^24 precision, but PHP allows more. I've updated the generator to ensure the test values do not exceed 2^24.

### DIFF
--- a/tests/Unit/Generation/TestNameValueProducerTest.php
+++ b/tests/Unit/Generation/TestNameValueProducerTest.php
@@ -33,7 +33,8 @@ final class TestNameValueProducerTest extends TestCase
     {
         $file = ProtoLoader::loadDescriptor('Utils/example.proto');
         // ProtoCatalog depends on the FileDescriptorProtos already being augmented.
-        ProtoAugmenter::Augment(Vector::new([$file]));
+        $files = Vector::new([$file]);
+        ProtoAugmenter::Augment($files);
         $catalog = new ProtoCatalog($files);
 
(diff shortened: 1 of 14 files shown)


noahdietz (approved):

tests/Unit/Generation/TestNameValueProducerTest.php
```diff
@@ -0,0 +1,49 @@
...
+use Google\Generator\Utils\ProtoAugmenter;
+use Google\Generator\Utils\ProtoCatalog;
+
+final class TestNameValueProducerTest extends TestCase
+{
+    public function testNoMoreThan24BitsOfPrecision(): void
+    {
+        $file = ProtoLoader::loadDescriptor('Utils/example.proto');
+        // ProtoCatalog depends on the FileDescriptorProtos already being augmented.
+        ProtoAugmenter::Augment(Vector::new([$file]));
```
noahdietz, line 36:
```suggestion
        $files = ProtoAugmenter::Augment(Vector::new([$file]));
```

bshaffer: I agree that this would be ideal, but the `ProtoAugmenter` currently returns `void`

noahdietz: ack

PR #591, opened 2023-03-08
chore: Rename optionalArgs to callOptions

saranshdhingra:
Although we didn't need to rename the var `$optionalArgs` in the generator files, but it seemed consistent to do so.


src/Generation/GapicClientGenerator.php
```diff
@@ -48,6 +48,10 @@
 
 class GapicClientGenerator
 {
+    // This is the variable name for the callOptions(prev optionalArgs)
+    // in the GAPIC clients.
+    const CALL_OPTIONS_VAR = 'callOptions';
```
bshaffer, line 53: nit: make this a private constant
```suggestion
    private const CALL_OPTIONS_VAR = 'callOptions';
```

bshaffer (commented):
Looks like the tests are still failing. I'm not sure how you updated the tests here, but I ran `php tests/Unit/ProtoTests/GoldenUpdateMain.php` and then supplied `0` for all tests, and PHPUnit passed.

saranshdhingra:
> Looks like the tests are still failing. I'm not sure how you updated the tests here, but I ran php tests/Unit/ProtoTests/GoldenUpdateMain.php and then supplied 0 for all tests, and PHPUnit passed.

Yeah, I did the same, I also ran the `bazel run` commands.
When I run locally on 8.1 without `USE_TOOLS_PROTOC=true` and it does show some errors which are on `v2-dev` as well 🤔 

Trying to debug it.

PR #592, opened 2023-03-08
fix(bazel): exclude fragment files from package

noahdietz:
This is the easiest/dirtiest way to exclude the generated fragment `.build.txt` files. It also makes it easier to debug later on b.c the fragments aren't deleted from the archive generated by `php_gapic_library`.

The "post-processor" version of this would be to `rm` the fragment file after applying it to the protobuf-php file, but I'm not sure we want to remove it, as this will make future debugging a tad more difficult. Open to suggestions though e.g. only delete the fragment if it was successfully applied.


bshaffer (approved):
LGTM!

PR #593, opened 2023-03-09
fix(v2): final empty classes, internal base classes

noahdietz:
Make the `v2` empty "service" clients `final` and the `v2` `BaseClient` classes `@internal`.

Fixes http://b/271286629

### DIFF
--- a/tests/Integration/goldens/iam/src/V1/Client/BaseClient/IAMPolicyBaseClient.php
+++ b/tests/Integration/goldens/iam/src/V1/Client/BaseClient/IAMPolicyBaseClient.php
@@ -68,6 +68,8 @@ use GuzzleHttp\Promise\PromiseInterface;
  * This class provides the ability to make remote calls to the backing service through method
  * calls that map to API methods.
  *
+ * @internal
+ *
  * @method PromiseInterface getIamPolicyAsync(GetIamPolicyRequest $request, array $optionalArgs = [])
  * @method PromiseInterface setIamPolicyAsync(SetIamPolicyRequest $request, array $optionalArgs = [])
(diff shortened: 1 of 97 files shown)


bshaffer (approved):

tests/Integration/goldens/iam/src/V1/Client/BaseClient/IAMPolicyBaseClient.php
```diff
@@ -68,6 +68,8 @@
  * This class provides the ability to make remote calls to the backing service through method
  * calls that map to API methods.
  *
+ * @internal
```
bshaffer, line 71: Hmm, I just had a thought. I think if we label something as internal it's skipped by the breaking change detector :-(

But i don't know another way we can say "don't use this class directly" that wouldn't affect the check.

bshaffer: Hopefully it's smart enough to see the breaking changes in the non-internal surfaces.  I'll have to test it out

noahdietz: I wonder if the Empty Client extending the BaseClient would give the BCD a non-internal surface to vet...

bshaffer: I just checked, and the BC breaking client throws an error if it finds a non-internal class extending that parent class!

So my fears are invalid, and everything is great.

noahdietz: Woo! thanks for raising and for checking :)

PR #596, opened 2023-03-13
chore: Rename optionalArgs to callOptions

saranshdhingra:


saranshdhingra:
So the bazel tests are passing locally, could it just be the cache in play?
Also, @noahdietz I didn't use the command before opening the PRs, what options do I have now?

noahdietz:
> So the bazel tests are passing locally, could it just be the cache in play? Also, @noahdietz I didn't use the command before opening the PRs, what options do I have now?

Rotate the cache value, then rerun the GitHub Checks manually.

noahdietz (commented):
LGTM but let's let @bshaffer review as well.

bshaffer (approved):
LGTM with one suggestion

PR #598, opened 2023-03-13
chore: Add bazel helper scripts

saranshdhingra:


bshaffer (commented):
Question: why is this on the v2 branch? this applies to main as well.

Also, nit, but I _think_ the `scripts` dir is for bazel targets. Maybe this would be better off in `tests/scripts`?

saranshdhingra:
> Question: why is this on the v2 branch? this applies to main as well.

I thought the same, but I would like to use this while we work on v2. If I merged this in `main`, then we'd have to sync main into `v2-dev`, which may not always be the case? If that is completely fine, then I can take that approach and resend the PR with `main` as the base.

> Also, nit, but I _think_ the `scripts` dir is for bazel targets. Maybe this would be better off in `tests/scripts`?

Sure. Although, my thought was that this would be similar to the `dev/sh` folder in `google-cloud-php` where we have helper scripts, so others might come later. But `tests/scripts` sounds good too :)

bshaffer:
>  If I merged this in main, then we'd have to sync main into v2-dev
Yes! We keep this branch up-to-date with main

scripts/run_bazel_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+# Usage: ./run_bazel_tests.sh
+
+# This runs the bazel test commands, ex:
+# bazel test //tests/Integration:asset && \
+# bazel test //tests/Integration:compute_small
+# etc
+
+echo "Running bazel tests"
+
+bazel run //tests/Integration:asset_update && \
```
noahdietz, line 12:
```suggestion
bazel test //tests/Integration:asset && \
```

Same goes for all other targets in **this** file.

noahdietz: Not quite. `bazel test` just runs the test cases. `bazel run ...:{api}_update` will actually update the golden files for the integration tests but not run a test. These are two different things.

saranshdhingra: Oops. I see my mistake now. Thanks for pointing it out.

PR #599, opened 2023-03-14
feat: Make constants private in GapicV2Generator

saranshdhingra:
I have 2 questions here @noahdietz @bshaffer :

1. We might need to access these constants somewhere in the code, should I add a static getter for these constants? And if I end up doing that, then what is the point of making it private 🤔 
2. We also need to change these variables sometimes for development(like staging endpoints etc when an environment variable is present), should that be done in this PR?

### DIFF
--- a/src/Generation/GapicClientV2Generator.php
+++ b/src/Generation/GapicClientV2Generator.php
@@ -146,6 +146,7 @@ class GapicClientV2Generator
     {
         return AST::constant('SERVICE_NAME')
             ->withPhpDocText('The name of the service.')
+            ->withAccess(Access::PRIVATE)
             ->withValue($this->serviceDetails->serviceName);
     }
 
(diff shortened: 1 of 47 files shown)


noahdietz:
> > We might need to access these constants somewhere in the code, should I add a static getter for these constants? And if I end up doing that, then what is the point of making it private 🤔
> 
> If we end up needing to access these constants elsewhere in the code, then we'd want to make them protected or public. But I do not believe that will be the case.

Yeah, let's make it private for now, then see if a use case comes along that we didn't think of, and address it. Easier to loosen things later than it is to tighten it.

> > We also need to change these variables sometimes for development
> 
> That's possible through client options, and I do not believe that behavior needs to change from V1

Indeed, most of these e.g. SERVICE_ADDRESS are "default" values. Configurable from client options during construction.

noahdietz (approved):

PR #601, opened 2023-03-16
chore: Make GAPIC client classes abstract

saranshdhingra:


src/Ast/PhpClass.php
```diff
@@ -28,13 +28,18 @@ final class PhpClass extends AST
 {
     use HasPhpDoc;
 
-    public function __construct(Type $type, ?ResolvedType $extends, bool $final)
-    {
+    public function __construct(
+        Type $type,
+        ?ResolvedType $extends,
+        bool $final,
```
vishwarajanand, line 34: Should this be by default `false`?

vishwarajanand (approved):

src/Ast/PhpClass.php
saranshdhingra: Thought so too. But,
1. It defaults to `false` in `AST.php`
2. I didn't wanna change it before confirming.

Perhaps @bshaffer can suggest

bshaffer: It's required, so I think that's better than having a default

src/Ast/PhpClass.php
```diff
@@ -98,7 +109,14 @@ public function withMembers(Vector $members): PhpClass
     public function toCode(): string
     {
         $extends = is_null($this->extends) ? '' : " extends {$this->extends->toCode()}";
-        $class = $this->final ? 'final class' : 'class';
+
+        $class = 'class';
+        if($this->final) {
+            $class = 'final class';
+        } elseif( $this->abstract) {
```
bshaffer, line 116: We should add a CS checker, as one would have caught this! (spaces after `if` and `elseif`, no extra space to start conditional)
```suggestion
        if ($this->final) {
            $class = 'final class';
        } elseif ($this->abstract) {
```

src/Ast/PhpClass.php
```diff
@@ -98,7 +109,14 @@ public function withMembers(Vector $members): PhpClass
     public function toCode(): string
     {
         $extends = is_null($this->extends) ? '' : " extends {$this->extends->toCode()}";
-        $class = $this->final ? 'final class' : 'class';
+
+        $class = 'class';
```
bshaffer, line 113: nit: let's throw an exception if they're both set
```suggestion
        $class = 'class';
        if ($this->final && $this->abstract) {
            throw new RuntimeException('Cannot use the final modifier on an abstract class');
        }
```

src/Generation/GapicClientV2Generator.php
```diff
@@ -98,7 +98,11 @@ private function generateImpl(): PhpFile
 
     private function generateClass(): PhpClass
     {
-        return AST::class($this->serviceDetails->gapicClientV2Type)
+        return AST::class(
+                $this->serviceDetails->gapicClientV2Type,
+                null,
+                false,
+                true)
```
bshaffer, line 105: Let's use PHP8.1 named arguments!
```suggestion
        return AST::class(
                $this->serviceDetails->gapicClientV2Type,
                abstract: true)
```

bshaffer (commented):
a few suggestions but this is looking great!

src/Ast/PhpClass.php
noahdietz: The reason that the constructor has a required param is because the main entry point for `PhpClass` creation is the `AST::class` method, which has optional params. The *only* instance of `new PhpClass` is in `AST::class`. As such, I'd prefer the constructor params to be required for the time being, so please change `bool $abstract` to required.

> We could also consider making this an enum:

I like this idea, though I think the parameter for it would need to be an array of `ClassModifier`s(?) so that it could be [`Static`, `Final`] for example. We'd need to also validate which options can be used together of course.

src/Ast/PhpClass.php
noahdietz: +1

src/Generation/GapicClientV2Generator.php
noahdietz: I _think_ the generator supports being run as `7.4` and `8.0` still per: https://github.com/googleapis/gapic-generator-php/blob/main/composer.json#L30

Not sure if that is valid still or not. I'd be happy to bump the PHP version of this to `^8.1` though.

src/Ast/PhpClass.php
bshaffer: That makes sense! I think the only class modifiers are abstract and final, and can't be used together. But yes, an enum array would also work! In a separate PR would be better though.

noahdietz: > in a separate PR would be better though.

Yep, for sure.

src/Generation/GapicClientV2Generator.php
bshaffer: The `composer.json` for the V2 branch says that it requires 8.1. I'm not sure if this is accurate or requires more work, though.

noahdietz: Oh yes yes. Then we should be good to go with your suggestion! My b

saranshdhingra: Good idea. It completely skipped my mind that we're  using `8.1`.

src/Ast/PhpClass.php
saranshdhingra: 👍 Will tackle the enums part in another PR.

PR #603, opened 2023-03-21
chore: various fixes from PoC review

noahdietz:
* Fix spacing in fragments
* update gax-php dependency
* add `@experimental` to `build` methods
* remove `registerTemplates` helper that was moved to gax-php
* add simple comment to `__call`
* add type hints to v2 client resource name helpers


src/Generation/BuildMethodFragmentGenerator.php
```diff
@@ -48,7 +48,7 @@ public static function format(string $code): string
         // Remove the wrapping class
         preg_match(sprintf('/' . $codeWrap . '/s', '(.*)'), $code, $matches);
 
-        return $matches[1] . PHP_EOL;
+        return PHP_EOL . $matches[1] . PHP_EOL;
```
bshaffer, line 51: It's a bit picky but... the PRs are generating two new lines after every `build` function (see [here](https://github.com/googleapis/google-cloud-php/pull/5970/files#diff-959739e5002dd513ece7f65b869b98a9f0b5064f0e6230337d00eb4aa1351267R63-R64)), so I _think_ we want to remove the second `PHP_EOL`
```suggestion
        return PHP_EOL . $matches[1];
```

noahdietz: Cool, yeah it's because it there is already a line break after where it is injected. I'll make this change with the updated baselines.

noahdietz: Done

bshaffer (approved):

PR #608, opened 2023-04-06
feat(v2): add PRE_MIGRATION_SURFACE_ONLY option

noahdietz:
Adds new mode `PRE_MIGRATION_SURFACE_ONLY` to basically keep the v1 generator code generation mode intact. This will allow PA clients not managed by us in google-cloud-php to adopt the new v2 surface (and the gax upgrades along with it) when they are ready.

Tested against SecretManager via local owlbot pipeline. No diffs were generated in the existing v1 client when using the new mode.


tests/Unit/ProtoTests/ProtoTest.php
```diff
@@ -122,7 +122,7 @@ public function testDeprecatedService(): void
 
     public function testBasicDiregapic(): void
     {
-        $this->runProtoTest('BasicDiregapic/library_rest.proto', 'google.example.library.v1', 'rest');
+        $this->runProtoTest('BasicDiregapic/library_rest.proto', 'google.example.library.v1', 'rest', migrationMode: MigrationMode::PRE_MIGRATION_SURFACE_ONLY);
```
bshaffer, line 125: This is for sure nits, but maybe we should rename these tests based on the migration mode, just `testMigrationModePreMigrationSurfaeOnly()` and so on and so forth?

src/Generation/ResourcesGenerator.php
```diff
@@ -43,15 +44,18 @@ private static function ensureDecimal(string $s): string
...
                     $descriptor = [
-                        'callType' => AST::literal('\Google\ApiCore\Call::LONGRUNNING_CALL'),
-                        // Include the responseType for Custom Operations because they define their own operation class.
-                        'responseType' => $method->responseType->getFullName(true),
                         'longRunning' => static::customOperationDescriptor($serviceDetails, $method)
                     ];
+                    if (!$preMigrationOnly) {
+                        // Include the responseType for Custom Operations because they define their own operation class.
+                        $descriptor['responseType'] = $method->responseType->getFullName(true);
+                        $descriptor['callType'] = AST::literal('\Google\ApiCore\Call::LONGRUNNING_CALL');
```
bshaffer, line 57: I feel like you probably told me this but, is there any reason why "PRE_MIGRATION_SURFACE_ONLY" clients can't use these descriptors? They seem to be a V1 or V2 feature. Is it just because it's a change that will cause a lot of churn?

bshaffer:
The bazel build is failing on:

```
//tests/Integration:speech                                               FAILED in 0.0s
  /github/home/.cache/bazel/_bazel_root/b4a2200717111fd40b6d4b376224a710/execroot/gapic_generator_php/bazel-out/k8-fastbuild/testlogs/tests/Integration/speech/test.log
```

Just an FYI, I', not sure why that would be happening

src/Generation/ResourcesGenerator.php
noahdietz: Not all of the values are available in the older versions of GAX (requiring a GAX upgrade, complicating rollout of this stuff) and yes to avoid churn/extraneous config that isn't used by the client code at all.

tests/Unit/ProtoTests/ProtoTest.php
noahdietz: Sure thing. I named them `test{OriginalName}WithPreMigrationSurface` and so on, because ultimately we don't want to lose that the test was originally for, and because we will remove `migration_mode` in generator v2 (once v1 is dropped).

bshaffer: thanks! that makes sense

bshaffer (approved):

noahdietz:
> The bazel build is failing on:
> 
> ```
> //tests/Integration:speech                                               FAILED in 0.0s
>   /github/home/.cache/bazel/_bazel_root/b4a2200717111fd40b6d4b376224a710/execroot/gapic_generator_php/bazel-out/k8-fastbuild/testlogs/tests/Integration/speech/test.log
> ```
> 
> Just an FYI, I', not sure why that would be happening

Figured it out - needed to update the `{target}_update` script to delete the `php.txt` fragment files as well when updating. In this case, Speech being moved back to pre-migration surface, wasn't generating fragment files anymore, but running `speech_update` didn't remove the old ones that had been checked in. So I updated the bazel script implementation to also remove those, `speech_update` did the right thing, and it works now.

PR #610, opened 2023-04-11
fix(v2): ignore mixin methods during fragment gen

noahdietz:
Mixin methods don't have their request/response messages as part of the `php_gapic_assembly_pkg`, so they aren't able to be `PostProcessor`'d, and thus cannot have a build method fragment generated for them. We could potentially look at fixing this later, but for now, this is the most reasonable solution.

Tested via googleapis build that failed before, but passes now.


bshaffer (approved):
Surprised this didn't update any goldens 🤔

Issue #624, opened 2023-05-08
error thrown when building ContainerAnalysis with migration_mode MIGRATING


bshaffer:
See [the failed build here](https://fusion2.corp.google.com/invocations/7e1d60b1-5c8e-4b9a-aaaf-8effd2f0ab5e/targets/actools%2Fgoogleapis%2Fstable%2Finternal%2Fpresubmit/log)

When building the PHP library for `google/devtools/containeranalysis/v1`, the following error is thrown:

```
Warning: file_get_contents(/bazel_root/1d081bf08a15278ca34ce779797e8450/sandbox/processwrapper-sandbox/10063/execroot/com_google_googleapis/bazel-out/k8-fastbuild/bin/google/devtools/containeranalysis/v1/google-cloud-devtools-containeranalysis-v1-php/google-cloud-devtools-containeranalysis-v1-php/proto/src/Google/Cloud/Iam/V1/TestIamPermissionsRequest.php): Failed to open stream: No such file or directory in /bazel_root/1d081bf08a15278ca34ce779797e8450/execroot/com_google_googleapis/bazel-out/k8-opt-exec-2B5CBBC6/bin/external/gapic_generator_php/rules_php_gapic/out/install/src/PostProcessor/FragmentInjectionProcessor.php on line 52

Fatal error: Uncaught TypeError: Google\PostProcessor\FragmentInjectionProcessor::__construct(): Argument #1 ($contents) must be of type string, bool given, called in /bazel_root/1d081bf08a15278ca34ce779797e8450/execroot/com_google_googleapis/bazel-out/k8-opt-exec-2B5CBBC6/bin/external/gapic_generator_php/rules_php_gapic/out/install/src/PostProcessor/FragmentInjectionProcessor.php on line 53 and defined in /bazel_root/1d081bf08a15278ca34ce779797e8450/execroot/com_google_googleapis/bazel-out/k8-opt-exec-2B5CBBC6/bin/external/gapic_generator_php/rules_php_gapic/out/install/src/PostProcessor/FragmentInjectionProcessor.php:82
```

It looks like the post processing step is trying to write a fragment to a non-existent Iam request. It's possible that the fragment inection processor is finding an annotation on a mixin and trying to write fragments for it.

bshaffer:
[...]

PR #626, opened 2023-05-08
fix: do not generate fragments for requests in other packages

bshaffer:
fixes https://github.com/googleapis/gapic-generator-php/issues/624

Better solution than #625


noahdietz (approved):
Well done

PR #629, opened 2023-05-10
fix: remove oneof wrappers in v2

### LINKED ISSUE #628: bug: "OneOf" GAPIC class is used in tests instead of protobuf oneof
In the generated GAPIC tests, the `Oneof` class is used inappropriately in tests by being passed to the `Request` instead of the protobuf message.

An example of this issue can be seen in the new [DiscoveryEngine](https://github.com/googleapis/google-cloud-php/pull/6212) client in the test [`UserEventServiceClientTest::importUserEventsTest`](https://github.com/googleapis/google-cloud-php/pull/6212/files#diff-8c9c7a216f57ce40c1a6eeeb3f22cba1e2c35eadfb6f20eef2b0ac291f02f143R178-R184): 

```php
        // Mock request
        $source = new SourceOneof();
        $source->setInlineSource(new InlineSource());
        $formattedParent = $gapicClient->dataStoreName('[PROJECT]', '[LOCATION]', '[DATA_STORE]');
        $request = (new ImportUserEventsRequest())
            ->setInlineSource($source)
            ->setParent($formattedParent);
```

This throws [the following error](https://github.com/googleapis/google-cloud-php/actions/runs/4939064015/jobs/8829451896?pr=6212#step:8:241):

```
1) Google\Cloud\DiscoveryEngine\Tests\Unit\V1beta\Client\UserEventServiceClientTest::importUserEventsTest
Exception: Expect Google\Cloud\DiscoveryEngine\V1beta\ImportUserEventsRequest\InlineSource.

google-cloud-php/vendor/google/protobuf/src/Google/Protobuf/Internal/GPBUtil.php:198
google-cloud-php/DiscoveryEngine/src/V1beta/ImportUserEventsRequest.php:84
google-cloud-php/DiscoveryEngine/tests/Unit/V1beta/Client/UserEventServiceClientTest.php:183
phpvfscomposer://google-cloud-php/vendor/phpunit/phpunit/phpunit:106
```

The solution I found in the test is to pass the `InlineSource` protobuf message directly into the request, instead of the `Oneof` class:

```php
        // Mock request
        // $source = new SourceOneof();
        // $source->setInlineSource(new InlineSource());
        $source = new InlineSource();
        $formattedParent = $gapicClient->dataStoreName('[PROJECT]', '[LOCATION]', '[DATA_STORE]');
        $request = (new ImportUserEventsRequest())
            ->setInlineSource($source)
            ->setParent($formattedParent);
```

Some questions I have are:

1. Is this new behavior? I know other GAPIC libraries have `Oneof` messages, so why hasn't this happened before?
2. What is the purpose of the `SourceOneof` class? It does not seem to be generated by `protobuf`, but by the GAPIC generator. But protobuf handles `Oneof` types at a message level, so what does the `SourceOneof` class do?
3. Is this an easy fix and I should just stop asking so many questions?

bshaffer:
fixes #628

 - Removes use of OneOf wrapper classes in V2 tests
 - Removes use of OneOf wrapper classes in V2 samples
 - Skips generating OneOf wrapper classes for `MigrationMode::NEW_SURFACE_ONLY`


src/Generation/TestNameValueProducer.php
```diff
@@ -145,7 +147,8 @@ public function fieldInit(MethodDetails $method, FieldDetails $field, &$fieldVar
 
         // This should only use oneof wrapper types if the oneof is on the top level request message.
         $inTopLevel = $method->inputMsg === $field->containingMessage;
-        if ($field->isOneOf && $inTopLevel) {
+        // This should only be used in V1 tests, as V2 does not need Oneof wrappers.
+        if ($field->isOneOf && $inTopLevel && !$this->v2) {
```
noahdietz, line 151: I don't think we need this `$v2` variable, we can check the `$method->serviceDetails->migrationMode`

bshaffer: The logic isn't dependent on Migration Mode, it depends on whether or not you're _using_ the V1 or V2 client. Since some migration modes have both, the only way to know is to explicitly pass it in.

I changed the variable name to `$useOneofWrappers` instead but I'm not sure that's better. I am definitely open to any better ways to do this.

bshaffer: I could potentially change it to `$isV2Test`... I think that might make it the most clear

noahdietz: Oh ok thanks!

noahdietz (approved):

PR #631, opened 2023-05-25
fix: increase php memory limit

bshaffer:
Increase PHP memory limit for compute + v2 surface, as the build is currently crashing because it's running out of memory

Should allow builds in https://github.com/googleapis/googleapis/pull/807 to pass!


noahdietz (approved):

bshaffer:
Getting same unrelated failures in https://github.com/googleapis/gapic-generator-php/pull/631. Maybe some dependency updated to cause this.

bshaffer:
Tests are fixed in https://github.com/googleapis/gapic-generator-php/pull/632

PR #635, opened 2023-05-31
fix: include non-required resource fields in helpers

noahdietz:
Previously, the algorithm for identifying resources in scope of helper generation only traversed fields of type message that were annotated as required (and of course, represented a resource). This resulted in some such fields being excluded from helper generation, when they are in fact part of the request message tree.

This just removes that check on `isRequired` field, bringing such references in scope.

Updates #453


bshaffer (approved):
LGTM

PR #639, opened 2023-06-21
chore: rename processor interface

bshaffer:
I missed this the first time around, but all interfaces should be suffixed with `Interface`. This is not an official PHP standard, but it is standard across all the `googleapis` libraries we maintain.


noahdietz (approved):

noahdietz:
Need to change the `PostProcessor` import and typing I think

bshaffer:
Hmm, they were just flakey and needed to be rerun 💀

PR #647, opened 2023-08-25
fix: add optional-to-required fix for analytics admin

bshaffer:
Fix optional field which was changed to required in AnalyticsAdmin

See https://github.com/googleapis/googleapis/blob/master/google/analytics/admin/v1beta/analytics_admin.proto#L275 See https://github.com/googleapis/google-cloud-php/pull/6588

**NOTE**: This API is NOT GA, so it may be best to just take the breaking change

### DIFF
--- a/src/Generation/FieldDetails.php
+++ b/src/Generation/FieldDetails.php
@@ -138,6 +138,7 @@ class FieldDetails
      * AFTER a package's 1.0 release, back to being optional.
      */
     private static $optionalToRequiredFixes = [
+        'google.analytics.admin.v1beta.UpdateMeasurementProtocolSecretRequest' => ['updateMask'],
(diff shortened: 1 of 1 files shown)


noahdietz (approved):

noahdietz:
Can't wait for this to go away in the major bump

bshaffer:
Do you think it's best to just take the breaking change in this case? I'm on the fence...

noahdietz:
It might...Breaking change as non major version (since its beta)?

bshaffer:
Yes, exactly, and the library itself isn't GA (although it's on [v0.17.0](https://github.com/googleapis/php-analytics-admin/releases/tag/v0.17.0), which is pretty dang stable!

noahdietz:
Yeah makes sense to me to just take the break, if that's what you want to do (post Next)

bshaffer:
Closing because we have decided to take the pre-GA breaking change

PR #649, opened 2023-09-05
fix: optional-to-required field for Cloud DMS

bshaffer:
Fix BC-breaking changes in https://github.com/googleapis/google-cloud-php/pull/6607

Also organizes array values alphabetically for readability

### DIFF
--- a/src/Generation/FieldDetails.php
+++ b/src/Generation/FieldDetails.php
@@ -138,15 +138,16 @@ class FieldDetails
      * AFTER a package's 1.0 release, back to being optional.
      */
     private static $optionalToRequiredFixes = [
-        'google.logging.v2.UpdateCmekSettingsRequest' => ['name', 'cmek_settings'],
-        'google.logging.v2.GetCmekSettingsRequest' => ['name'],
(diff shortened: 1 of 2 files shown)


src/Generation/FieldDetails.php
```diff
@@ -138,15 +138,16 @@ class FieldDetails
      * AFTER a package's 1.0 release, back to being optional.
      */
     private static $optionalToRequiredFixes = [
-        'google.logging.v2.UpdateCmekSettingsRequest' => ['name', 'cmek_settings'],
-        'google.logging.v2.GetCmekSettingsRequest' => ['name'],
+        'google.cloud.clouddms.v1.DescribeDatabaseEntitiesRequest' => ['tree'],
```
vishwarajanand, line 141: What is the `tree` part and where is it used?

vishwarajanand (approved):
Approving to unblock, but left a Q for my understanding.

src/Generation/FieldDetails.php
bshaffer: I don't know anything about it other than it's a property that used to be optional and now it's required.

PR #658, opened 2023-09-26
feat: add support for Universe Domain via service address template

bshaffer:


src/Generation/GapicClientGenerator.php
```diff
@@ -155,10 +156,31 @@ private function serviceName(): PhpClassMember
...
+                $this->hasServiceAddressTemplate()
+                    ? PhpDoc::deprecated('SERVICE_ADDRESS_TEMPLATE should be used instead.')
+                    : null
+            ))
             ->withValue($this->serviceDetails->defaultHost);
     }
 
+    private function hasServiceAddressTemplate(): bool
+    {
+        return str_contains($this->serviceDetails->defaultHost, '.googleapis.com');
```
vishwarajanand, line 170:
```suggestion
        return !str_contains($this->serviceDetails->defaultHost, '.googleapis.com');
```

vishwarajanand: I think the name implies this to be a not contains check.
Like if default host does NOT contain googleapis.com, it has a service address template.

bshaffer: A service address template is the default host, but with `googleapis.com` replaced by `UNIVERSE_DOMAIN`. So if "googleapis.com" doesn't exist in the template, there's no place for us to put `UNIVERSE_DOMAIN`, so we can't give it a service address template.

Would it be more clear if we renamed the function something like `canGenerateServiceAddressTemplate`?

vishwarajanand: my bad, this seems correct

src/Generation/GapicClientGenerator.php
```diff
@@ -131,6 +131,7 @@ private function generateClass(): PhpClass
             ->withTrait($this->ctx->type(Type::fromName(\Google\ApiCore\GapicClientTrait::class)))
             ->withMember($this->serviceName())
             ->withMember($this->serviceAddress())
+            ->withMembers(Vector::new($this->hasServiceAddressTemplate() ? [$this->serviceAddressTemplate()] : []))
```
vishwarajanand, line 134: Why do we have a `Vector` here?

bshaffer: I did it because it may not exist, but looks like we can just supply `null` to `withMember` and achieve the same thing.
```suggestion
            ->withMember($this->hasServiceAddressTemplate() ? $this->serviceAddressTemplate() : null)
```

src/Generation/GapicClientGenerator.php
bshaffer: It may be clearer to call it `shouldHaveServiceAddressTemplate`?

src/Generation/GapicClientV2Generator.php
```diff
@@ -136,6 +136,7 @@ private function generateClass(): PhpClass
                 $this->serviceDetails->hasResources ? $this->ctx->type(Type::fromName(\Google\ApiCore\ResourceHelperTrait::class)): null)
             ->withMember($this->serviceName())
             ->withMember($this->serviceAddress())
+            ->withMembers(Vector::new($this->hasServiceAddressTemplate() ? [$this->serviceAddressTemplate()] : []))
```
bshaffer, line 139:
```suggestion
            ->withMember($this->hasServiceAddressTemplate() ? $this->serviceAddressTemplate() : null)
```

vishwarajanand (approved):
Lgtm

src/Generation/GapicClientGenerator.php
bshaffer: Thanks for the suggestion!!

PR #659, opened 2023-09-27
feat: consolidate BaseClient classes in V2 generator

bshaffer:
This will require
 - [ ] an update to `google/gax` to ensure that the method `GapicClientTrait::isNewClientSurface()` works as expected for the consolidated v2 clients (see https://github.com/googleapis/gax-php/pull/493)
 - [ ] an update to `google/cloud` to ensure that the consolidated v2 clients require the latest version of `google/gax` with the above feature.


noahdietz (approved):

noahdietz:
Nicely done! Love to see code reduction :)

PR #669, opened 2023-11-03
fix: prevent duplicate fragment gen

noahdietz:
Prevents duplicate generation of fragments for RPCs that share a request message (not common, but possible).


bshaffer (approved):

bshaffer:
If it's not too difficult, I'd love to see a test for this!

noahdietz:
> If it's not too difficult, I'd love to see a test for this!

We don't have any unit tests for this code, but there is one integration test. I will add one with the API that this fix is for once it is in googleapis, if that is OK.

noahdietz:
> > If it's not too difficult, I'd love to see a test for this!
> 
> We don't have any unit tests for this code, but there is one integration test. I will add one with the API that this fix is for once it is in googleapis, if that is OK.

fwiw I did test it locally against the offending API

PR #686, opened 2024-02-19
feat: Enable UUID auto-generation for fields listed for auto population in service.yaml

yash30201:
See b/318607731

~Files changed count would decrease drastically once https://github.com/googleapis/gapic-generator-php/pull/689 gets merged to main and main gets merged to this PR.~
Files changed count has decreased drastically as https://github.com/googleapis/gapic-generator-php/pull/689 has been  merged to main and this PR has been updated with main.

gax-php pr to implement the auto population logic: https://github.com/googleapis/gax-php/pull/543


src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,56 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+                }
+            }
+        }
+        $autoPopulatedFields = $autoPopulatedFields->map(fn ($x) => $x->camelName)->toArray();
+
+        // Convert array of field_names to associative array of
+        // field_name => field_type
+        return array_combine(
+            $autoPopulatedFields,
+            array_fill(0, count($autoPopulatedFields), 'UUID4')
```
yash30201, line 565: Since we would be the only consumers((in the gax)) of the "value" corresponding to each auto populated field name, I feel we can just hard a value here (`UUID4` in our case). We just need to make sure that this value is respected in GAX.

src/Generation/FieldDetails.php
```diff
@@ -236,6 +243,15 @@ private function determineIsRequired(DescriptorProto $containingMessage, FieldDe
         return $isRequired;
     }
 
+    private function isFieldInfoUuid(FieldDescriptorProto $field)
```
noahdietz, line 246: Nit: I'd name this `isFormatUuid4` - that's the specific field that contains this, a little more specific.

src/Generation/FieldDetails.php
```diff
@@ -236,6 +243,15 @@ private function determineIsRequired(DescriptorProto $containingMessage, FieldDe
...
+    {
+        $fieldInfo = ProtoHelpers::fieldInfo($field)->firstOrNull();
+        if ($fieldInfo) {
+            return $fieldInfo->getFormat() === 1;
```
noahdietz, line 250: Please use the enum in the comparison rather than the magic number, I think its `Google\Api\FieldInfo\Format::UUID4`.

src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,56 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+     * The field name is listed in `google.api.publishing.method_settings.auto_populated_fields`.
+     * The field is annotated with `google.api.field_info.format = UUID4`.
+     *
+     * @return array<string, string> Fields to autopopulate in GAX if unset by user
+     */
+    private static function getUuid4FieldsToAutoPopulate(
+        MethodDetails $method,
+        ServiceDetails $serviceDetails
+    ): array {
+        $autoPopulatedFields = Vector::new([]);
```
noahdietz, line 531:
```suggestion
        $autoPopulatedFields = Vector::new();
```
defaults to `[]`

src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,56 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+     * For a given method and service, returns any fields that are available
+     * for autopopulation given the restrictions below.
+     * The field is a top-level string field of a unary method's request message.
+     * The field is not annotated with `google.api.field_behavior = REQUIRED`.
+     * The field name is listed in `google.api.publishing.method_settings.auto_populated_fields`.
+     * The field is annotated with `google.api.field_info.format = UUID4`.
+     *
+     * @return array<string, string> Fields to autopopulate in GAX if unset by user
+     */
+    private static function getUuid4FieldsToAutoPopulate(
```
noahdietz, line 527: Nit: I'd name this in passive tense for "what it is" -> `getAutoPopulatedUuid4Fields`

src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,56 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+        MethodDetails $method,
+        ServiceDetails $serviceDetails
+    ): array {
+        $autoPopulatedFields = Vector::new([]);
+
+        $methodSettings = $serviceDetails->serviceYamlConfig->methodSettings;
+        foreach ($methodSettings as $methodSetting) {
+            $autoPopulatedFieldNames = iterator_to_array($methodSetting->getAutoPopulatedFields());
```
noahdietz, line 535: Can we leverage the `Vector`APIs and use the `filter`/`any` APIs to do this stuff? First of all, the `isStreaming` check should be at the top of the entire function - we don't need to find do this entire function if the `$method` given is streaming. Then, use `any` to check if the `$methodSettings` contains a setting for this method && has non-empty `autoPopulatedFields`, if no return early. If yes, then `foreach` until you find it then `break`. Then use `Vector` apis on the `$autoPopulatedFieldNames` to do all of the filtering on behavior and format, then map to camelName then to array.

src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,56 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+
+                // Check if the $methodSetting is for this particular $method
+                $splitMethodName = explode('.', $methodSetting->getSelector());
+                $methodName = $splitMethodName[array_key_last($splitMethodName)];
+                if ($methodName !== $method->name) {
+                    continue;
+                }
+
+                // Process only if the method is unary
+                if (!$method->isStreaming()) {
```
noahdietz, line 546: This check should be at the start of the method, exit as early as possible since we know the `$method` from the outset.

src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,56 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+                    $autoPopulatedFields = $method->optionalFields
+                        ->filter(
+                            // Filter out field names to autopopulate from all fields of the method
+                            fn ($f) => in_array($f->name, $autoPopulatedFieldNames)
+                        )
+                        ->filter(
+                            // Filter for optional field and uuid4 format
+                            fn ($f) => !$f->isRequired && $f->isUuid4
+                        );
```
noahdietz, line 555: I'm confused with this bit: Is it filtering through optional fields that appear in the auto populated list, and then checks if those are required? That seems like duplicate work?

src/Generation/ResourcesGenerator.php
noahdietz, line 565: You can use the `Vector` `toMap` function to do this `field name => UUID4` mapping I think

src/Generation/ResourcesGenerator.php
yash30201: Updated.

src/Generation/FieldDetails.php
yash30201: Yeah this looks better. Updated it.

src/Generation/FieldDetails.php
yash30201: Makes sense. I've updated it.

src/Generation/ResourcesGenerator.php
yash30201: Yep makes sense. Thanks

src/Generation/ResourcesGenerator.php
yash30201: Ahh Snap!
I forgot removing that `!$f->isRequired`. I've removed it now.

This happened because I was using `$method->alllFields` earlier and implemented the logic. Then later found that we can simply use `$method->optionalFields`

src/Generation/ResourcesGenerator.php
yash30201: Yes you're correct. I've utilised the Map Api now.

src/Generation/ResourcesGenerator.php
```diff
@@ -524,45 +524,33 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+                $splitName = explode('.', $x->getSelector());
+                return $method->name == $splitName[array_key_last($splitName)];
+            })->firstOrNull();
+
+        if (!is_null($methodSetting) && count($methodSetting->getAutoPopulatedFields())) {
+            $fieldNames = iterator_to_array($methodSetting->getAutoPopulatedFields());
+            $result = $method->optionalFields
+                ->filter(fn ($x) => in_array($x->name, $fieldNames))
+                ->filter(fn ($x) => $x->isUuid4)
+                ->toMap(fn ($x) => $x->camelName, fn ($x) => 'UUID4');
```
noahdietz, line 548: I think you mentioned this in an earlier thread, but should we use the enum constant directly here instead of a string? We use the constant for call type here: https://github.com/googleapis/gapic-generator-php/blob/7cd894f4812f1ffdeeb4b5f35390ca9b71c395c2/src/Generation/ResourcesGenerator.php#L57

src/Generation/ResourcesGenerator.php
```diff
@@ -524,45 +524,33 @@ private static function getQueryParams(MethodDetails $method): Vector
...
         }
-        $autoPopulatedFields = $autoPopulatedFields->map(fn ($x) => $x->camelName)->toArray();
 
-        // Convert array of field_names to associative array of
-        // field_name => field_type
         return array_combine(
-            $autoPopulatedFields,
-            array_fill(0, count($autoPopulatedFields), 'UUID4')
+            $result->keys()->toArray(),
+            $result->values()->toArray()
```
noahdietz, line 553: This feels like something we should have in the `Map` API e.g. `toArray()` which just returns the underlying `$data` property (maybe a copy of it) which is itself an associative array...WDYT? Would you mind adding/using that?

src/Generation/ResourcesGenerator.php
```diff
@@ -524,45 +524,33 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+            return [];
+        }
+
+        $methodSetting = $service->serviceYamlConfig->methodSettings
+            ->filter(function ($x) use ($method) {
+                $splitName = explode('.', $x->getSelector());
+                return $method->name == $splitName[array_key_last($splitName)];
+            })->firstOrNull();
+
+        if (!is_null($methodSetting) && count($methodSetting->getAutoPopulatedFields())) {
```
noahdietz, line 543: Nit: again, perhaps "exit on unfulfilled condition, fallthrough to green path" refactor so

```php
if (is_null(...) || !count(...)) {
  return [];
}

// rest of code 
```

src/Generation/ResourcesGenerator.php
yash30201: Yeah I totally support this feeling. I was skeptical on calling this out because the Vector, Map, Set...these API feels like they mimic C++ style data structures to some extent and hence their design is kind of same for all languages (I may be wrong here as I never verified this fact).

Anyways, now that you also called out, I've extracted the a `toArray()` code to `Map` api.

src/Generation/ResourcesGenerator.php
yash30201: Sure. Updated the code.

src/Generation/ResourcesGenerator.php
yash30201: Oh, thanks for the info.
Done

src/Generation/ResourcesGenerator.php
```diff
@@ -507,4 +513,46 @@ private static function getQueryParams(MethodDetails $method): Vector
...
+     * The field is annotated with `google.api.field_info.format = UUID4`.
+     *
+     * @return array<string, string> Fields to autopopulate in GAX if unset by user
+     */
+    private static function getAutoPopulatedUuid4Fields(
+        MethodDetails $method,
+        ServiceDetails $service
+    ): array {
+        $result = [];
+
```
bshaffer, line 532: nit: `$result` variable does nothing
```suggestion
```

src/Utils/ServiceYamlConfig.php
```diff
@@ -81,6 +82,11 @@ public function __construct(?string $serviceYaml)
             if (!is_null($apis)) {
                 $this->apiNames = Vector::new($apis)->map(fn ($a) => $a->getName());
             }
+
+            $publishing = $service->getPublishing();
+            if (!is_null($publishing)) {
```
bshaffer, line 87: nit, but why not just
```suggestion
            if ($publishing = $service->getPublishing()) {
```

src/Collections/Map.php
```diff
@@ -266,4 +266,17 @@ public function get($key, $default)
         [$exists, $value] = static::apply($this->data, $key, 0, null);
         return $exists ? $value : $default;
     }
+
+    /**
+     * Returns a copy of Map as PHP associative array.
+     *
+     * @return array<mixed, mixed>
+     */
+    public function toArray()
```
bshaffer, line 275: typehint
```suggestion
    public function toArray(): array
```

tests/Unit/ProtoTests/ProtoTest.php
```diff
@@ -150,4 +150,8 @@ public function testDisableSnippetGeneration(): void
             false
         );
     }
+
+    public function testBasicAutoPopulation(): void {
```
bshaffer, line 154: PHP code standards - no cuddly braces for class methods
```suggestion
    public function testBasicAutoPopulation(): void
    {
```

tests/Unit/ProtoTests/GoldenUpdateMain.php
```diff
@@ -148,6 +152,6 @@ private static function updateGolden(int $testIndex)
 }
 
 if (isset($argv)) {
-    $selection = $argv[1] ?? -1;
+    $selection = (int)($argv[1] ?? -1);
```
bshaffer, line 155: code standards
```suggestion
    $selection = (int) ($argv[1] ?? -1);
```

src/Collections/Map.php
bshaffer, line 275: Do we want to be extra explicit and call this `toAssociateArray`?

noahdietz: I'd support that, yeah.

noahdietz (approved):
Looks great @yash30201 well done. Resolve @bshaffer 's comments and we can merge this.

src/Utils/ServiceYamlConfig.php
yash30201: Verified that the default value is `null`. Updated.

src/Generation/ResourcesGenerator.php
yash30201: Done

src/Collections/Map.php
yash30201: Same thought. I named it `toArray` so that it's similar to Vector API. But since you folks also think this should be `toAssociativeArray`, then why not! Updated.

bshaffer: yes that was a typo, I meant `toAssociativeArray` (or `toAssocArray`). Thank you!

bshaffer (approved):

PR #687, opened 2024-02-22
feat(Spanner): Support Emulator via env in Spanner Admin Client

ajupazhamayil:
This PR introduces emulator support in the Spanner Admin Client via an environment variable. These changes will simplify future similar requirements.


tests/Unit/ProtoTests/BasicDiregapic/out/src/Gapic/LibraryGapicClient.php
```diff
@@ -1272,7 +1272,7 @@ public function findRelatedBooks($names, $shelves, array $optionalArgs = [])
      *     $operationResponse->pollUntilComplete();
      *     if ($operationResponse->operationSucceeded()) {
      *         $result = $operationResponse->getResult();
-     *     // doSomethingWith($result)
+     *         // doSomethingWith($result)
```
ajupazhamayil, line 1275: I changed these here just to resolve some unittests issues.

vishwarajanand:
Test failures look unrelated, potentially should get fixed by: https://github.com/googleapis/gapic-generator-php/pull/688

src/Generation/EmulatorSupportGenerator.php
```diff
@@ -0,0 +1,89 @@
...
+{
+    /**
+     * Map of emulator support required clients and their expected env variables.
+     */
+    private static $emulatorSupportFixes = [
+        '\Google\Cloud\Spanner\Admin\Database\V1\Client\DatabaseAdminClient' => 'SPANNER_EMULATOR_HOST',
+        '\Google\Cloud\Spanner\Admin\Instance\V1\Client\InstanceAdminClient' => 'SPANNER_EMULATOR_HOST',
+        // Added for unittesting
+        '\Testing\Basic\Client\BasicClient' => 'BASIC_EMULATOR_HOST'
+    ];
```
vishwarajanand, line 38: Is this supposed to work only for Spanner? I think this way could be used to generate the emulator vars for Datastore and others as well?

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +211,28 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
...
+    {
+        $emulatorHost = getenv('BASIC_EMULATOR_HOST');
+        if (!empty($emulatorHost)) {
+            if (parse_url($emulatorHost, PHP_URL_SCHEME) === $scheme) {
+                $search = $scheme . '://';
+                $emulatorHost = str_replace($search, '', $emulatorHost);
+            }
+
+            return [
+                'apiEndpoint' => $emulatorHost,
```
vishwarajanand, line 226: I think we are planning to move away from `apiEndpoint`.
I am not sure whether we should have `serviceAddress` or some other field instead.

src/Generation/EmulatorSupportGenerator.php
saranshdhingra: For now, we only want it to work on Spanner, but yes we may expand it to others later.

ajupazhamayil:
Referred: https://github.com/googleapis/gapic-generator-php/pull/688

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +211,30 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
...
+                'transportConfig' => [
+                    'grpc' => [
+                        'stubOpts' => [
+                            'credentials' => ChannelCredentials::createInsecure(),
+                        ],
+                    ],
+                ],
+                'credentials' => new InsecureCredentials(),
```
ajupazhamayil, line 234: TODO: Change this to a working InsecureCredentialsWrapper. I could only find the wrapper in the Core!

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: I see `apiEndpoint` used in the Gapic generated Admin clients here:             'apiEndpoint' => self::SERVICE_ADDRESS . ':' . self::DEFAULT_SERVICE_PORT,

tests/Tools/PhpClassComparer.php
```diff
@@ -45,7 +45,7 @@ class PhpClassComparer
      */
     public static function compare(string $phpClassOne, string $phpClassTwo, bool $printDiffs = true): bool
     {
-        $phpParser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+        $phpParser = (new ParserFactory)->createForNewestSupportedVersion();
```
ajupazhamayil, line 48: I changed these here just to resolve some unittests issues.

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
bshaffer: Using `apiEndpoint` is the correct option. The deprecated field is `serviceAddress`

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +211,30 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
...
     }
+
+    /** Configure the gapic configuration to use a service emulator. */
+    private function setEmulatorConfig()
+    {
+        $emulatorHost = getenv('BASIC_EMULATOR_HOST');
+        if (!empty($emulatorHost)) {
+            if ($scheme = parse_url($emulatorHost, PHP_URL_SCHEME)) {
+                $search = $scheme . '://';
+                $emulatorHost = str_replace($search, '', $emulatorHost);
```
bshaffer, line 222: This logic doesn't make sense to me - if someone provides a URL scheme in `BASIC_EMULATOR_HOST`, we strip it out? Why?

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +211,30 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
     {
         return $this->startApiCall('MethodWithArgs', $request, $callOptions)->wait();
     }
+
+    /** Configure the gapic configuration to use a service emulator. */
+    private function setEmulatorConfig()
```
bshaffer, line 216: this is not a setter, it's a getter (e.g. it returns the default Emulator config)

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -144,6 +146,7 @@ public function __construct(array $options = [])
     {
         $clientOptions = $this->buildClientOptions($options);
         $this->setClientOptions($clientOptions);
+        $options = array_merge($options, $this->setEmulatorConfig($options));
```
bshaffer, line 149: We should merge the emulator config options in _before_ calling `setClientOptions`. I would even suggest doing it before `buildClientOptions`:
```suggestion
        $clientOptions = $this->buildClientOptions($options + $this->getDefaultEmulatorConfig()); 
        $this->setClientOptions($clientOptions);
```

bshaffer, line 149: The main difference between `array_merge` and using `+` is that `array_merge` will merge `null` values even if the array key exists, whereas the `+` will not (if the array key exists in the first array, the second array key value is not merged in). IMHO, I think the second (using `+`) is the behavior we want, as it would allow users to override the default emulator config, e.g.:


```php
putenv('SPANNER_EMULATOR_HOST=myhost');
$spanner = new SpannerClient(['credentials' => null]); // the default credentials will be used instead of insecure ones
```

It's a very minor difference, and I can't really see anyone needing to do the above. But to me, this at least reflects what behavior I'd expect as a user.

src/Ast/AST.php
```diff
@@ -80,6 +80,21 @@ abstract class AST
...
     public const E_USER_ERROR = "\0E_USER_ERROR";
 
+    /** @var string Constant to reference `parse_url`. */
+    public const PARSE_URL = "\0parse_url";
+
+    /** @var string Constant to reference `str_replace`. */
+    public const STRING_REPLACE = "\0str_replace";
+
+    /** @var string Constant to reference `createInsecure`. */
+    public const CREATE_INSECURE = "\0createInsecure";
```
bshaffer, line 90: We are typically using these constants for PHP structures. Since this is just a function name called on a class, we should leave it out of the constants here, and just hardcode it when we need it

src/Generation/EmulatorSupportGenerator.php
```diff
@@ -0,0 +1,90 @@
...
+                    )),
+                AST::return(
+                    AST::array([
+                        'apiEndpoint' => $emulatorHostVar,
+                        'transportConfig' => [
+                            'grpc' => [
+                                'stubOpts' => [
+                                    'credentials' => AST::staticCall(
+                                        $ctx->type((Type::fromName("Grpc\ChannelCredentials"))),
+                                        AST::CREATE_INSECURE)()
```
bshaffer, line 69: just hardcode this
```suggestion
                                        AST::method('createInsecure'))()
```

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +211,30 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
...
+                        'stubOpts' => [
+                            'credentials' => ChannelCredentials::createInsecure(),
+                        ],
+                    ],
+                ],
+                'credentials' => new InsecureCredentials(),
+            ];
+        }
+
+        return [];
```
bshaffer, line 238: This is a nit, but best practice is to return early for something like this. So rather than have `if (!empty($emulatorHost))`, with the interesting logic inside a conditional, move the `return [];` to the top like this:

```php
private function getDefaultEmulatorConfig()
{
    $emulatorHost = getenv('BASIC_EMULATOR_HOST');
    if (empty($emulatorHost)) {
        return [];
    }

    // return the default config
}

File: src/Generation/EmulatorSupportGenerator.php [hunk 12]
```diff
@@ -0,0 +1,90 @@
...
+use Google\Generator\Ast\PhpDoc;
+use Google\Generator\Utils\Type;
+use \Grpc\ChannelCredentials;
+
+class EmulatorSupportGenerator
+{
+    /**
+     * Map of emulator support required clients and their expected env variables.
+     */
+    private static $emulatorSupportFixes = [
```
bshaffer, line 33: Don't call this "Fixes". We are not fixing anything here, we are adding a feature (emulator support). Would be better to just call it something like `emulatorSupportClients`, since the keys are client names.

ajupazhamayil: Sounds great.

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: Sounds good

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: I have reused the logic from this core trait: https://github.com/googleapis/google-cloud-php/blob/9b40c2cc7e67d101ca0432a29c613b7c48053b4f/Core/src/EmulatorTrait.php#L31

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: Thank you Brent for the clarification

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: sounds good

ajupazhamayil:
@bshaffer Thank you for the quick review, I have resolved your comments.

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -31,6 +31,8 @@
 use Google\ApiCore\Transport\TransportInterface;
 use Google\ApiCore\ValidationException;
 use Google\Auth\FetchAuthTokenInterface;
+use Google\Cloud\Core\InsecureCredentialsWrapper;
```
ajupazhamayil, line 34: This is available only in the Core. @bshaffer WDYT? Should we move this class to gax?

bshaffer: Yes, I think that's a good idea. We can then deprecate this one, and put the new one in the `Google\ApiCore` namespace

ajupazhamayil:
I have checked the following test to understand it is in fact working for the Spanner Admin Client.

```
$diff tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php ../google-cloud-php/Spanner/src/Admin/Database/V1/Client/DatabaseAdminClient.php

<  * Copyright 2024 Google LLC
---
>  * Copyright 2023 Google LLC
38d37
< use Google\Cloud\Core\InsecureCredentialsWrapper;
71d69
< use Grpc\ChannelCredentials;
375c373
<         $clientOptions = $this->buildClientOptions($options + $this->getDefaultEmulatorConfig());
---
>         $clientOptions = $this->buildClientOptions($options);
1037,1062d1034
<     }
< 
<     /** Configure the gapic configuration to use a service emulator. */
<     private function getDefaultEmulatorConfig(): array
<     {
<         $emulatorHost = getenv('SPANNER_EMULATOR_HOST');
<         if (empty($emulatorHost)) {
<             return [];
<         }
< 
<         if ($scheme = parse_url($emulatorHost, PHP_URL_SCHEME)) {
<             $search = $scheme . '://';
<             $emulatorHost = str_replace($search, '', $emulatorHost);
<         }
< 
<         return [
<             'apiEndpoint' => $emulatorHost,
<             'transportConfig' => [
<                 'grpc' => [
<                     'stubOpts' => [
<                         'credentials' => ChannelCredentials::createInsecure(),
<                     ],
<                 ],
<             ],
<             'credentials' => new InsecureCredentialsWrapper(),
<         ];
```

src/Generation/EmulatorSupportGenerator.php
```diff
@@ -0,0 +1,87 @@
...
+use Google\Generator\Ast\PhpDoc;
+use Google\Generator\Utils\Type;
+
+class EmulatorSupportGenerator
+{
+    /**
+     * Map of emulator support required clients and their expected env variables.
+     */
+    private static $emulatorSupportClients = [
+        'Google\Cloud\Spanner\Admin\Database\V1\Client\DatabaseAdminClient' => 'SPANNER_EMULATOR_HOST',
```
bshaffer, line 32: What is the logic behind using a specific env var for different products? Why not just support a `GOOGLE_CLOUD_EMULATOR_HOST` which can work across products?

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -142,7 +144,7 @@ private static function getClientDefaults()
      */
     public function __construct(array $options = [])
     {
-        $clientOptions = $this->buildClientOptions($options);
+        $clientOptions = $this->buildClientOptions($options + $this->getDefaultEmulatorConfig());
```
bshaffer, line 147: we probably also want to update the documentation of the client object somewhere to include information about the environment variable, e.g. in the class's PHPDoc or the __constructor PHPDoc, something like:

```
Setting the "SPANNER_EMULATOR_HOST" environment variable will automatically set the API Endpoint to
the value specified in the variable, as well as ensure that empty credentials are used in the transport layer.
```

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
bshaffer: SGTM

src/Generation/EmulatorSupportGenerator.php
ajupazhamayil: This page specifically name it like `SPANNER_EMULATOR_HOST` : https://cloud.google.com/spanner/docs/emulator

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +210,30 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
     {
         return $this->startApiCall('MethodWithArgs', $request, $callOptions)->wait();
     }
+
+    /** Configure the gapic configuration to use a service emulator. */
+    private function getDefaultEmulatorConfig(): array
```
bshaffer, line 215: What if we moved the `$emulatorHost` parameter up into the method signature? Then we could move the entire method into `GapicClientTrait` instead of generating it every time:
```suggestion
    private function getDefaultEmulatorConfig(string $emulatorHost): array
```

src/Generation/EmulatorSupportGenerator.php
bshaffer: I know it's the officially supported one, I was just thinking that it's configured the same way for every product, so why have specific product names for it? But That's out of the scope of this PR.

tests/Tools/PhpClassComparer.php
ajupazhamayil: Merged it with main, this is resolved due to the latest PR. So, reverted this change.

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: I am open to the above mentioned idea although I like this approach.

I am thinking of going with the current implemented approach for now (as its a `private` method) and if we get a similar requirement from other products, we will look into multiple options and take the decision?

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: Done

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: Updated this, gax needs new release for this though

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
bshaffer: Yes that sounds great. I agree what we have here is better, but wanted to explore every option.

ajupazhamayil: Thank you!

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
bshaffer: This has been done in https://github.com/googleapis/google-cloud-php/pull/7089!

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -27,10 +27,12 @@
 use Google\ApiCore\ApiException;
 use Google\ApiCore\CredentialsWrapper;
 use Google\ApiCore\GapicClientTrait;
+use Google\ApiCore\InsecureCredentialsWrapper;
```
bshaffer, line 30: We need to remember to update these clients to have a minimum version of `google/gax:^1.29` before we roll any of these out. Otherwise, this will throw a class not found error in versions of GAX < 1.29.

I think the best way to handle this may be to just update ALL minimum versions of GAX to 1.29 now - or at the very least, update `Spanner/composer.json`.

bshaffer (commented):
One minor suggestion, otherwise this is looking great. Would also love some eyeballs from @noahdietz

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
ajupazhamayil: Updated the min gax version to 1.29.1. PR: https://github.com/googleapis/google-cloud-php/pull/7095

src/Generation/EmulatorSupportGenerator.php
```diff
@@ -0,0 +1,95 @@
...
+                        ],
+                        'credentials' => AST::new($ctx->type((Type::fromName("Google\ApiCore\InsecureCredentialsWrapper"))))(),
+                    ]))), AST::return(AST::array([])))
+            ->withPhpDoc(PhpDoc::block(
+                PhpDoc::text("Configure the gapic configuration to use a service emulator."),
+            ))
+            ->withReturnType($ctx->type(Type::array()));
+        } else {
+            return null;
+        }
```
noahdietz, line 79: Invert this, return early if not configured. Exiting early when a condition isn't met is easier to read top-to-bottom, and dedents the code (easier on the eyes)
```suggestion
        if (!array_key_exists($fullClassName, self::$emulatorSupportClients)) {
            return null;
        }

        return AST::method(self::DEFAULT_EMULATOR_CONFIG_FN)
            ->withAccess(Access::PRIVATE)
            ->withBody(AST::block(
                AST::assign($emulatorHostVar, AST::call(AST::GET_ENV)(self::$emulatorSupportClients[$fullClassName])),
                AST::if(AST::call(AST::EMPTY)($emulatorHostVar))->then(AST::return(AST::array([]))),
                AST::if(AST::assign($schemeVar, AST::call(AST::PARSE_URL)($emulatorHostVar, $phpUrlSchemeConst)))
                    ->then(AST::block(
                        AST::assign($searchVar, AST::binaryOp($schemeVar, '.', '://')),
                        AST::assign($emulatorHostVar, AST::call(AST::STRING_REPLACE)($searchVar, '', $emulatorHostVar)),
                    )),
                AST::return(
                    AST::array([
                        'apiEndpoint' => $emulatorHostVar,
                        'transportConfig' => [
                            'grpc' => [
                                'stubOpts' => [
                                    'credentials' => AST::staticCall(
                                        $ctx->type((Type::fromName("Grpc\ChannelCredentials"))),
                                        AST::method('createInsecure'))()
                                ]
                            ]
                        ],
                        'credentials' => AST::new($ctx->type((Type::fromName("Google\ApiCore\InsecureCredentialsWrapper"))))(),
                    ]))), AST::return(AST::array([])))
            ->withPhpDoc(PhpDoc::block(
                PhpDoc::text("Configure the gapic configuration to use a service emulator."),
            ))
            ->withReturnType($ctx->type(Type::array()));
```

You will need to fix tabbing from suggestion, or just edit yourself.

src/Generation/GapicClientV2Generator.php
```diff
@@ -560,7 +561,9 @@ private function construct(): PhpClassMember
             ->withParams($optionsParam)
             ->withAccess(Access::PUBLIC)
             ->withBody(AST::block(
-                Ast::assign($clientOptions, AST::call(AST::THIS, $buildClientOptions)($options)),
+                Ast::assign($clientOptions, AST::call(AST::THIS, $buildClientOptions)(
+                    EmulatorSupportGenerator::generateEmulatorOptionsIfRequired($this->serviceDetails, $options)
+                )),
```
noahdietz, line 566: I would make this simpler/easier to interpret by making it a separate line such that the generated code only _adds_ a line rather than _modifies_ an existing one. The generated code could instead be

```php
$clientOptions = $this->buildClientOptions($options); // existing, unchanged
$clientOptions = $clientOptions + $this->getDefaultEmulatorConfig(); // new, only added when configured
```

You may need to add the `Ast::assign($clientOptions...` and such to the `generateEmulatorOptions` function so its completely standalone.

src/Generation/EmulatorSupportGenerator.php
```diff
@@ -0,0 +1,95 @@
...
+{
+    /**
+     * Map of emulator support required clients and their expected env variables.
+     */
+    private static $emulatorSupportClients = [
+        '\Google\Cloud\Spanner\Admin\Database\V1\Client\DatabaseAdminClient' => 'SPANNER_EMULATOR_HOST',
+        '\Google\Cloud\Spanner\Admin\Instance\V1\Client\InstanceAdminClient' => 'SPANNER_EMULATOR_HOST',
+        // Added for unittesting
+        '\Testing\Basic\Client\BasicClient' => 'BASIC_EMULATOR_HOST'
+    ];
```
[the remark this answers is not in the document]
ajupazhamayil: Yeah, I think a PLC to consolidate all languages with the `ClientLibrarySettings` would be great. Will add to the backlog for now

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
```diff
@@ -208,4 +214,30 @@ public function methodWithArgs(RequestWithArgs $request, array $callOptions = []
...
+            'transportConfig' => [
+                'grpc' => [
+                    'stubOpts' => [
+                        'credentials' => ChannelCredentials::createInsecure(),
+                    ],
+                ],
+            ],
+            'credentials' => new InsecureCredentialsWrapper(),
+        ];
+    }
```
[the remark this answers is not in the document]
ajupazhamayil: We have discussed this idea before actually. With the current information about the problem, we thought both options have its pros and cons. So, we moved forward with the current implemented approach for now (as its a private method) thinking if we get a similar/improved requirement from other products, we will look into multiple options and take a decision. But I am open to change if you see this is a necessary change. Thank you!

src/Generation/EmulatorSupportGenerator.php
ajupazhamayil: Done

src/Generation/EmulatorSupportGenerator.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /** @var string Name of the default emulator config function. */
+    public const DEFAULT_EMULATOR_CONFIG_FN = 'getDefaultEmulatorConfig';
+
+    public static function generateEmulatorSupportIfRequired(ServiceDetails $serviceDetails, SourceFileContext $ctx) {
```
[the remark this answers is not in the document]
ajupazhamayil: Done

src/Generation/GapicClientV2Generator.php
ajupazhamayil: Sounds good, Changed, thank you!

src/Generation/EmulatorSupportGenerator.php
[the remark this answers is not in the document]
noahdietz: We (Cloud SDK) would add the appropriate config to the `ClientLibrarySettings`, then the service team only needs to update their Service config yaml with the property we define. That way, all generators consuming service config yaml can see that XYZ service clients should respect ABC emulator host env var.

No proto changes for service team, just one client library only config change.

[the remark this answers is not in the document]
ajupazhamayil: Great, I see this is a non blocking comment, we can discuss this offline and drive this then.. Thank you!

tests/Integration/BUILD.bazel
```diff
@@ -598,7 +598,7 @@ php_gapic_library(
     name = "spanner_php_gapic",
     migration_mode = "MIGRATING",
     srcs = ["@com_google_googleapis//google/spanner/admin/database/v1:database_proto_with_info"],
-    service_yaml = "@com_google_googleapis//google/spanner/admin/database/v1:spanner_admin_database.yaml",
+    service_yaml = "@com_google_googleapis//google/spanner/admin/database/v1:spanner_admin_database_gapic.yaml",
```
noahdietz, line 601: service yaml and gapic yaml are different things, you'll want to use [`spanner.yaml` instead](https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner.yaml)

noahdietz (approved):
One comment on the integration test but otherwise, LGTM

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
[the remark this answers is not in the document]
noahdietz: Fair enough! It can always be migrated to gax-php as you say...

tests/Integration/BUILD.bazel
ajupazhamayil: Thank you for this information, I was confused on what to actually add, I will change it right away.

ajupazhamayil: Changed. Thank you!

tests/Unit/ProtoTests/Basic/out/src/Client/BasicClient.php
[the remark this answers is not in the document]
bshaffer: I'm with @noahdietz that I think GAX is a better place for this, but also the work is done already and my main argument against it is just that we've already set the precedent for "thin generator, fat GAX" and this goes the other way. +1 that I'm okay with it as-is for now.

bshaffer:
@ajupazhamayil nit, but we only use tags like `feat(Spanner)` in `google-cloud-php`. It doesn't make sense in the generator here because this generator applies to all GAPIC APIs, and shouldn't be thought of as applying to a specific one. Even though this feature is only going to roll out to Spanner, it's a generic feature which could apply to any package/API.

ajupazhamayil:
> @ajupazhamayil nit, but we only use tags like `feat(Spanner)` in `google-cloud-php`. It doesn't make sense in the generator here because this generator applies to all GAPIC APIs, and shouldn't be thought of as applying to a specific one. Even though this feature is only going to roll out to Spanner, it's a generic feature which could apply to any package/API.

Sure @bshaffer, Will take care of this from next time onwards. Thank you for the information.

PR #688, opened 2024-02-22
fix: remove PhpFactory::create from PhpClassComparer

vishwarajanand:
Seems like `ParserFactory::create(..)` is deprecated: [ref](https://github.com/nikic/PHP-Parser/blob/af14fdb282aa0e288bfe7eb3b57893484b68dc27/UPGRADE-5.0.md#changes-to-the-parser-factory)

Also contains following changes to ensure a test pass:
1. `tests/Unit/ProtoTests/*/out/src/Gapic` -> these are files from which generated content is compared.
2. `tests/Integration/goldens` -> these are the files from which generated files are compared in integration tests.
3. `composer.json` -> needs a php version to be pinned for tests to run, else throws a dependent package min version mismatch errors.
4. `tests/Tools/PhpClassComparer.php` -> ensures we mitigate the deprecated old method in `ParserFactory`

### DIFF
--- a/.github/workflows/bazel_example.yml
+++ b/.github/workflows/bazel_example.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Build example package; all targets
       run: |
         cd bazel_example;
-        bazel build ...
+        bazel build //...
 
     - name: Extract Bazel package output
       run: |
--- a/.github/workflows/bazel_example.yml
+++ b/.github/workflows/bazel_example.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Build example package; all targets
       run: |
         cd bazel_example;
-        bazel build //...
+        bazel build ...
 
     - name: Extract Bazel package output
       run: |
--- a/tests/Tools/PhpClassComparer.php
+++ b/tests/Tools/PhpClassComparer.php
@@ -25,6 +25,7 @@ use PhpParser\Node;
 use PhpParser\NodeFinder;
 use PhpParser\Error;
 use PhpParser\ParserFactory;
+use PhpParser\PhpVersion;
 use PhpParser\Node\Stmt\ClassMethod;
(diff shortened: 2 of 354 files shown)


.github/workflows/bazel_example.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Build example package; all targets
       run: |
         cd bazel_example;
-        bazel build ...
+        bazel build //...
```
ajupazhamayil, line 49: Without this change also it works fine. I referred this change in my initial PR and there its working fine

vishwarajanand: The later is more powerful though in this context both would have same meaning

tests/Tools/PhpClassComparer.php
```diff
@@ -45,7 +45,7 @@ class PhpClassComparer
      */
     public static function compare(string $phpClassOne, string $phpClassTwo, bool $printDiffs = true): bool
     {
-        $phpParser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+        $phpParser = (new ParserFactory)->createForNewestSupportedVersion();
```
bshaffer, line 48: Wouldn't we want to use the lowest supported PHP version, e.g. PHP 7.4 in our case?

.github/workflows/bazel_example.yml
vishwarajanand: I meant `//...` searches for build targets in the entire workspace not just subdirectories.

vishwarajanand: reverted

tests/Tools/PhpClassComparer.php
vishwarajanand: Makes sense.. fixed it.

bshaffer (approved):

PR #692, opened 2024-02-28
fix: correct method for supportedTransports

bshaffer:
Woops! We've been writing the wrong method name all this time. The method is actually `supportedTransports`, but we've been generating `getSupportedTransports`. OH NO!

See [`ClientOptionsTrait::supportedTransports`](https://github.com/googleapis/gax-php/blob/54a863e63ee318308637adb283f6157ccc3aabbb/src/ClientOptionsTrait.php#L284) (called [here](https://github.com/googleapis/gax-php/blob/54a863e63ee318308637adb283f6157ccc3aabbb/src/ClientOptionsTrait.php#L103))

### DIFF
--- a/tests/Unit/ProtoTests/BasicDiregapic/out/src/Gapic/LibraryGapicClient.php
+++ b/tests/Unit/ProtoTests/BasicDiregapic/out/src/Gapic/LibraryGapicClient.php
@@ -229,8 +229,8 @@ class LibraryGapicClient
         return 'rest';
     }
 
-    /** Implements GapicClientTrait::getSupportedTransports. */
-    private static function getSupportedTransports()
+    /** Implements GapicClientTrait::supportedTransports. */
+    private static function supportedTransports()
     {
         return [
             'rest',
--- a/src/Generation/GapicClientGenerator.php
+++ b/src/Generation/GapicClientGenerator.php
@@ -139,7 +139,7 @@ class GapicClientGenerator
             ->withMember($this->operationsClient())
             ->withMember($this->getClientDefaults())
             ->withMember($this->defaultTransport())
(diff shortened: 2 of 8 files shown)


ajupazhamayil (approved):
Great catch. LGTM.

tests/Unit/ProtoTests/BasicDiregapic/out/src/Gapic/LibraryGapicClient.php
```diff
@@ -229,8 +229,8 @@ private static function defaultTransport()
         return 'rest';
     }
 
-    /** Implements GapicClientTrait::getSupportedTransports. */
-    private static function getSupportedTransports()
+    /** Implements GapicClientTrait::supportedTransports. */
```
ajupazhamayil, line 232: How did you catch this?

bshaffer: I was reviewing @yash30201 's doc on Storage V2, and how it only will support `grpc`, so I wanted to check what the method was that we'd need to generate (which is this one), and noticed the names were different between here and GAX.

bshaffer: ...it would be better if we had a test, though.

bshaffer: Test added!

https://github.com/googleapis/gapic-generator-php/pull/692/commits/d32d4ba1c6edcd9eaf75ba4ff219bbeb9490aeca

noahdietz (approved):

PR #696, opened 2024-03-22
fix: add license and generated code headers to generated resource files

### LINKED ISSUE #643: Ensure all generated files are labeled as generated in the comments
Namely, the PHP config files generated in the `resources/` directory, but we should try to add some identifying comment for all generated files.

Unfortunately, JSON does not support comments, so those are the exception.

Hectorhammett:
Change the `ResourcesGenerator` class to make use of the `AST::file()` method to add the required comments. I also added some comments for discussion on this PR to add them in a different PR if we come with a more elegant solution.

Fixes #643

### DIFF
--- a/src/Generation/ResourcesGenerator.php
+++ b/src/Generation/ResourcesGenerator.php
@@ -172,7 +172,7 @@ class ResourcesGenerator
                 ->toArray(fn($x) => $x->getNameCamelCase(), fn($x) => $x->getPattern());
         }
 
-        $return = AST::return(
+        $codeBlock = AST::return(
             AST::array([
                 'interfaces' => AST::array([
                     $serviceDetails->serviceName => AST::array($serviceDescriptor)
@@ -180,7 +180,13 @@ class ResourcesGenerator
             ])
         );
 
-        return "<?php\n\n{$return->toCode()};";
+        $currentYear = (int)date("Y");
+
+        return AST::file(null)
+            ->withApacheLicense($currentYear)
+            ->withGeneratedCodeWarning()
+            ->withBlock($codeBlock)
+            ->toCode() . ";"; 
     }
 
     public static function customOperationDescriptor(ServiceDetails $serviceDetails, MethodDetails $method)
@@ -276,10 +282,17 @@ class ResourcesGenerator
             $config['numericEnums'] = true;
         }
 
-        $return = AST::return(
+        $codeBlock = AST::return(
             AST::array($config)
         );
-        return "<?php\n\n{$return->toCode()};";
+
+        $currentYear = (int)date("Y");
+        
+        return AST::file(null)
+            ->withApacheLicense($currentYear)
+            ->withGeneratedCodeWarning()
+            ->withBlock($codeBlock)
+            ->toCode() . ';';
     }
 
     private static function compileRestConfigInterfaces(ServiceDetails $serviceDetails, ServiceYamlConfig $serviceYamlConfig)
--- a/src/Ast/PhpFile.php
+++ b/src/Ast/PhpFile.php
@@ -40,7 +40,7 @@ final class PhpFile extends AST
         return $this->clone(fn ($clone) => $clone->uses = $uses);
     }
 
-    public function withApacheLicense(int $year)
+    public function withApacheLicense(int $year): PhpFile
     {
         $license = <<<EOF
             /*
@@ -64,7 +64,7 @@ final class PhpFile extends AST
         return $this->clone(fn ($clone) => $clone->headerLines = $clone->headerLines->concat($license));
     }
 
-    public function withGeneratedCodeWarning()
+    public function withGeneratedCodeWarning(): PhpFile
     {
         $warning = <<<'EOF'
             /*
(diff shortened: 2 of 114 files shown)


src/Generation/ResourcesGenerator.php
```diff
@@ -172,15 +172,21 @@ public static function generateDescriptorConfig(ServiceDetails $serviceDetails,
...
         );
 
-        return "<?php\n\n{$return->toCode()};";
+        $currentYear = (int)date("Y");
+
+        return AST::file(null)
+            ->withApacheLicense($currentYear)
+            ->withGeneratedCodeWarning()
+            ->withBlock($codeBlock)
+            ->toCode() . ";";
```
Hectorhammett, line 189: I am adding the `;` manually here but I am not particularly fond of this. 

My first attempt was to add this to the `return` method on `AST` but that function is used in different places. I am using the `PhpFile` class which has the ability to use a codeblock but any `PhpBlock` class implements the `ShouldNotApplySemicolonInterface` which the `AST::toPhp` method checks to decided wether or not add a semicolon.

We could create a `EndPhpBlock` or `FinalPhpBlock` class that does not implements the interface perhaps?
```php
AST::file(null)
    ->withBlock(new EndPhpBlock($string))
```

 Add an optional parameter to force the semicolon?:
```php
AST::file(null)
    ->withBlock(new PhpBlock($string))
    ->withSemiColon()
```

perhaps a method on the `PhpClass` to "finalize" the class? But perhaps the word for "Final" might be confused by a Final class. Ready perhaps? that adds a `;` to the end?
```php
AST::file(null)
    ->withBlock(new PhpBlock())
    ->ready()
```

src/Generation/ResourcesGenerator.php
```diff
@@ -276,10 +282,17 @@ public static function generateRestConfig(ServiceDetails $serviceDetails, Servic
...
         );
-        return "<?php\n\n{$return->toCode()};";
+
+        $currentYear = (int)date("Y");
+        
+        return AST::file(null)
+            ->withApacheLicense($currentYear)
+            ->withGeneratedCodeWarning()
+            ->withBlock($codeBlock)
+            ->toCode() . ';';
```
Hectorhammett, line 295: Check above comment ☝️

src/Generation/ResourcesGenerator.php
noahdietz: Interesting...I don't want to mess with `PhpBlock` too much, especially if it is already opinionated about semicolons. Given how choosy the `AST` is about semicolons altogether, I think this might be a reasonable solution to keep it how it is...or do what you suggested with a new `EndPhpBlock`...maybe `LastPhpBlock`? It just wraps a `Block` and tacks on the `;` in `toCode`?

Hectorhammett: I poked around a bit more today to see if adding the extra PHP class would help but after debugging and poking further turns out that in this case, it is actually not using the `phpBlock` class, but just a regular `AST` class and the `toPhp` method just calls `toCode` on an AST instance:
```php
elseif ($x instanceof AST) {
    return $x->toCode();
}
```

So the semicolon interface does nothing in this case lol. I think it would be best to have a way to add a semicolon codewise but for now let's keep it like it is.

noahdietz: > So the semicolon interface does nothing in this case lol. I think it would be best to have a way to add a semicolon codewise but for now let's keep it like it is.

Nice code spelunking! I agree, simple enough for now. Maybe file a `priority: p3` `type: cleanup` bug to look at this?

noahdietz (commented):
You will also need to update the Bazel integration test Golden files. Those can be updated locally via the instructions in [DEVELOPMENT.md](https://github.com/googleapis/release-please/blob/main/README.md)

noahdietz:
When updating the integration test golden files you will need to rotate the bazel cache key in the secrets. It isn't really a "secret" just where this value is stored, which indicates we want a full rebuild of the bazel workspace.

`uuidgen | gh secret set CACHE_VERSION -r https://github.com/googleapis/gapic-generator-php`

This is documented here: https://github.com/googleapis/gapic-generator-php/blob/main/DEVELOPMENT.md#rotating-the-bazel-cache-key. You may or may not have perms for this, lmk if you don't.

I've rotated and rerun the build for you.

MODULE.bazel.lock
noahdietz, line 1: Please remove this file

MODULE.bazel
noahdietz, line 1: Please remove this file.

noahdietz (commented):
Please remove the MODULE bazel files - these are created by buildifier or something when you open a WORKSPACE or BUILD.bazel file. We don't support bzlmod yet.

Maybe we should add these to the .gitignore....

Hectorhammett:
> Please remove the MODULE bazel files - these are created by buildifier or something when you open a WORKSPACE or BUILD.bazel file. We don't support bzlmod yet.
> 
> Maybe we should add these to the .gitignore....

I removed these. Is there a reason why NOT to add them to the .gitignore? If not I could just add them rn unless there is a specific reason not to.

noahdietz:
> I removed these. Is there a reason why NOT to add them to the .gitignore? If not I could just add them rn unless there is a specific reason not to.

Not really, unless we worry about folks forgetting that we are ignoring them if we eventually add bzlmod support haha.

Let's just add it! This PR or another is fine.

Hectorhammett:
> > I removed these. Is there a reason why NOT to add them to the .gitignore? If not I could just add them rn unless there is a specific reason not to.
> 
> Not really, unless we worry about folks forgetting that we are ignoring them if we eventually add bzlmod support haha.
> 
> Let's just add it! This PR or another is fine.

@noahdietz 
Let's do it in a different PR. This is getting too mangled for what it is. Just a small code change and 114 files change feels like when you have a play dough getting mixed with different colours and ends up gray 🤣

noahdietz (approved):

PR #706, opened 2024-04-19
feat: Add support for the API version header

Hectorhammett:
Added support for the API version header from the new Versioning header for apis.

### DIFF
--- a/src/Ast/PhpProperty.php
+++ b/src/Ast/PhpProperty.php
@@ -18,13 +18,18 @@ declare(strict_types=1);
 
 namespace Google\Generator\Ast;
 
+use Google\Generator\Utils\ResolvedType;
+
 /** A property within a class. */
 final class PhpProperty extends PhpClassMember
 {
+    private ?ResolvedType $type;
+    
     public function __construct(string $name)
     {
         $this->name = $name;
         $this->value = null;
+        $this->type = null;
     }
 
     /**
@@ -39,6 +44,18 @@ final class PhpProperty extends PhpClassMember
         return $this->clone(fn ($clone) => $clone->value = $value);
     }
 
+    /**
+     * Adds a type declaration to the property
+     * 
+     * @param ResolvedType $type The type of the property
+     * 
+     * @return PhpProperty
+     */
+    public function withType(ResolvedType $type): PhpProperty
+    {
+        return $this->clone(fn ($clone) => $clone->type = $type);
+    } 
+
     public function getName(): string
     {
         return $this->name;
@@ -49,7 +66,13 @@ final class PhpProperty extends PhpClassMember
         return
             $this->phpDocToCode() .
             $this->accessToCode() .
+            $this->typeToCode() .
             '$' . $this->name .
             (is_null($this->value) ? ';' : ' = ' . static::toPhp($this->value) . ';');
+    }
+
+    private function typeToCode(): string
+    {
+        return $this->type ? $this->type->toCode() . ' ' : '';
     }
 }
--- a/src/Generation/GapicClientV2Generator.php
+++ b/src/Generation/GapicClientV2Generator.php
@@ -32,6 +32,7 @@ use Google\ApiCore\ValidationException;
 use Google\Auth\FetchAuthTokenInterface;
 use Google\Generator\Ast\AST;
 use Google\Generator\Ast\Access;
+use Google\Generator\Ast\TypeDeclaration;
 use Google\Generator\Ast\PhpClass;
 use Google\Generator\Ast\PhpClassMember;
 use Google\Generator\Ast\PhpDoc;
@@ -139,6 +140,7 @@ class GapicClientV2Generator
             ->withMember($this->hasServiceAddressTemplate() ? $this->serviceAddressTemplate() : null)
             ->withMember($this->servicePort())
             ->withMember($this->codegenName())
+            ->withMember($this->apiVersion())
             ->withMember($this->serviceScopes())
             ->withMember($this->operationsClient())
             ->withMember($this->getClientDefaults())
@@ -158,6 +160,19 @@ class GapicClientV2Generator
             ->withPhpDocText('The name of the service.')
             ->withAccess(Access::PRIVATE)
             ->withValue($this->serviceDetails->serviceName);
+    }
+
+    private function apiVersion()
+    {
+        if (is_null($this->serviceDetails->apiVersion)) {
+            return;
+        }
+
+        return AST::property('apiVersion')
+            ->withPhpDocText('The api version of the service')
+            ->withAccess(Access::PRIVATE)
+            ->withType(ResolvedType::string())
+            ->withValue(AST::literal("'" . $this->serviceDetails->apiVersion . "'"));
     }
 
     private function serviceAddress(): PhpClassMember
--- a/tests/Unit/ProtoTests/BasicGrpcOnly/out/src/Client/BasicGrpcOnlyClient.php
+++ b/tests/Unit/ProtoTests/BasicGrpcOnly/out/src/Client/BasicGrpcOnlyClient.php
@@ -44,7 +44,7 @@ final class BasicGrpcOnlyClient
     private const SERVICE_NAME = 'testing.basicgrpconly.BasicGrpcOnly';
 
     /** The default address of the service. */
-    private const SERVICE_ADDRESS = 'basicGrpcOnly.example.com';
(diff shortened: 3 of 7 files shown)


src/Generation/GapicClientV2Generator.php
```diff
@@ -160,6 +161,19 @@ private function serviceName(): PhpClassMember
...
+    private function apiVersion()
+    {
+        if (is_null($this->serviceDetails->apiVersion)) {
+            return;
+        }
+
+        return AST::property('apiVersion')
+            ->withPhpDocText('The api version of the service')
+            ->withAccess(Access::PRIVATE)
+            ->withType('string')
```
Hectorhammett, line 173: I gave this a thought and I don't love how I am just sending a string.

I was wondering that using something similar like the line above with `Access::PRIVATE` would look better, but I wonder how we can handle different classes, specially ones that we create. Of course this is with the thought that we should support any class in order to keep it generic.

Perhaps we could create an Enum with just the basic classes for our needs?

noahdietz: Like `AST::TYPE_STRING`, `AST::TYPE_ARRAY`, etc. ? Sure that makes sense to me. I agree "magic values" aren't as fun as "magic" might make them seem.

I wouldn't worry about adding constants for each known possible type just yet. Better to just add them as we need them.

Hectorhammett: Yeah something like that.

I am gonna add that then. I like the idea of using the types we need as we need them 👍

Hectorhammett: I ended up creating a class inside the `hasType` file to keep everything tidy as in the `Access` file. The class that contains all the constants is called `TypeDeclaration`.

noahdietz (approved):
This is tight tight hella tight. Well done Hector

src/Ast/HasType.php
```diff
@@ -0,0 +1,44 @@
...
+declare(strict_types=1);
+
+namespace Google\Generator\Ast;
+
+class TypeDeclaration {
+    /** @var string Constant to define a string type */
+    public const STRING = 'string';
+}
+
+trait HasType
```
bshaffer, line 26: This should be `HasTypeTrait`

src/Ast/HasType.php
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+declare(strict_types=1);
+
+namespace Google\Generator\Ast;
+
+class TypeDeclaration {
```
bshaffer, line 21: This should be in a separate class file, and we can add other constants to it there.

bshaffer (changes requested):

src/Ast/HasType.php
Hectorhammett: I am currently following the convention that is on the `PhpClassMember` file. It has 3 traits:
image

I would like to keep things nice and tidy. Do we vote for updating all the traits to call them `trait`?

src/Ast/HasType.php
Hectorhammett: Good catch! 

I updated my code to make use of `ResolvedType`. Should we remove this one then? I vote yes.

Hectorhammett: I decided to remove this one in favor for the `ResolvedType` type

src/Ast/HasType.php
bshaffer: Since the precedent has already been set, we can keep them named as-is. Thanks for pointing that out.

src/Ast/PhpClassMember.php
```diff
@@ -23,6 +23,7 @@ abstract class PhpClassMember extends AST
 {
     use HasPhpDoc;
     use HasAccess;
+    use HasType;
```
bshaffer, line 26: I would suggest moving this to just `PhpProperty`, as the other three classes which extend this class (`PhpConstant`, `PhpComment`, and `PhpMethod`) don't support types in the same way:

 - Constants: defining types are invalid (since they're implicit from the constant value)
 - Comments: don't have types
 - Methods: types are already defined by parameters return types

And we probably want to just get rid of the trait all together, since traits are meant for reusing code, and we will no longer be doing that.

Hectorhammett: Great comment. Makes sense, let me change it.

src/Ast/PhpProperty.php
```diff
@@ -18,13 +18,16 @@
...
+use Google\Generator\Utils\ResolvedType;
+
 /** A property within a class. */
 final class PhpProperty extends PhpClassMember
 {
     public function __construct(string $name)
     {
         $this->name = $name;
         $this->value = null;
+        $this->type = null;
```
bshaffer, line 30: As of PHP 8.2, [setting dynamic properties is deprecated](https://php.watch/versions/8.2/dynamic-properties-deprecated). This repo is currently on PHP 8.1, which is why no errors are thrown. We can wait until we update the repo to PHP 8.2 to add these properties, or we could add it now, e.g. 

```php
final class PhpProperty extends PhpClassMember
{
    private ?ResolvedType $type;
```

... but since we'll have to update these across the whole library eventually, I'll leave that up to you whether you'd like to define it now or not.

tests/Unit/ProtoTests/BasicGrpcOnly/out/src/Client/BasicGrpcOnlyClient.php
```diff
@@ -86,7 +86,7 @@ private static function supportedTransports()
      *
      *     @type string $apiEndpoint
      *           The address of the API remote host. May optionally include the port, formatted
-     *           as "<uri>:<port>". Default 'basicGrpcOnly.example.com:443'.
+     *           as "<uri>:<port>". Default 'basicgrpconly.example.com:443'.
```
bshaffer, line 89: woops 😦

bshaffer (approved):
Looks great! Just one suggestion

src/Ast/PhpProperty.php
Hectorhammett: Ehhh let's change it once and for all.

tests/Unit/ProtoTests/BasicGrpcOnly/out/src/Client/BasicGrpcOnlyClient.php
Hectorhammett: I rebased with main and this should be the correct name isn't it? The proto defines it with lower case. No idea why it changed now.

src/Ast/PhpProperty.php
bshaffer: If it's between introducing a "consistent" change that goes against a best practice, and introducing an "inconsistent" change that follows best practices, I'd prefer to go with the latter

tests/Unit/ProtoTests/BasicGrpcOnly/out/src/Client/BasicGrpcOnlyClient.php
bshaffer: I meant woops because I introduced this in the previous PR by changing it to lowercase before merging. This would have been fine, but I forgot to update the fixtures (as you've done here).

I'd say it's mostly the fault that the tests can pass even when they're accurate, because they aggressively cache.

src/Ast/PhpProperty.php
bshaffer: ...and I think having undefined properties is a really bad practice

Hectorhammett: Woops, I have this change but did not included the file on the commit. Will fix it rn.

Hectorhammett: Woops, I have this change but didn't include the file on the Commit. Fixing it now.

src/Ast/PhpProperty.php
```diff
@@ -18,13 +18,18 @@
 
 namespace Google\Generator\Ast;
 
+use Google\Generator\Utils\ResolvedType;
+
 /** A property within a class. */
 final class PhpProperty extends PhpClassMember
 {
+    private ?ResolvedType $type;
```
Hectorhammett, line 26: Added the property here, I resolved the conversation below.

Issue #705, opened 2024-04-17
Feature Request: Allow for gRPC-only GAPIC libraries


bshaffer:
In `php_gapic.bzl`, we throw an error when the transport omits `rest`:

https://github.com/googleapis/gapic-generator-php/blob/46dc31c550399de766d877038ec7d188f13fcfc9/rules_php_gapic/php_gapic.bzl#L73

There's no reason why we can't support gRPC-only GAPICs. In practice, this is already the case - the `Redis`, `Firestore`, and `Spanner` libraries, which all contain GAPIC components, require the gRPC-extension and so are essentially gRPC-only (although the user could provide a `REST` transport to them manually if they wanted to override this functionality).

Supporting gRPC-only libraries _shouldnt_ require a lot of work - we just need to ensure `supportedTransports` only returns `grpc` and `grpc-fallback` (similar to what we do in the [Compute clients](https://github.com/googleapis/google-cloud-php/blob/6da93edc3278a11d99fe9a8c23bb58456cecc543/Compute/src/V1/Client/DisksClient.php#L142-L148), but for gRPC instead of REST):

```php
/** Implements GapicClientTrait::supportedTransports. */
private static function supportedTransports()
{
    return ['grpc', 'grpc-fallback'];
}
```

At the moment, When we want to make a client "grpc-only" , we just require the gRPC extension in the package. But for clients where no REST http annotations exist (As is the case in See https://github.com/googleapis/google-cloud-php/pull/7245),  if someone manually configures the REST transport, this would throw a PHP fatal error. By specifying which transports are allowed at the client level, we make sure that the dependency on the grpc extension is a separate decision not tied to the client libraries, and make sure we provide a good error for customers who manually configure a different transport.

**NOTE**: `supportedTransports` has been moved from `GapicClientTrait` to `ClientOptionsTrait`, and we should update the phpdoc to reflect that

PR #707, opened 2024-04-23
feat: Add support for grpc only gapic

Hectorhammett:
fixes https://github.com/googleapis/gapic-generator-php/issues/705


rules_php_gapic/php_gapic.bzl
```diff
@@ -69,9 +69,7 @@ def php_gapic_srcjar(
     # Transport.
     if transport == None:
         transport = "grpc+rest"
-    if transport == "grpc":
-        fail("Error: gRPC-only PHP GAPIC libraries are not yet supported")
-    if transport != "grpc+rest" and transport != "rest":
+    if transport != "grpc+rest" and transport != "rest" and transport != "grpc":
         fail("Error: Only 'grpc+rest' or 'rest' transports are supported")
```
Hectorhammett, line 73: What error would be here? 
```
"Error: Only 'grpc+rest', 'rest' or `grpc` transports are supported"
```

Or more like an `invalid` error?

src/Generation/GapicClientV2Generator.php
```diff
@@ -478,15 +478,23 @@ private function defaultTransport()
 
     private function supportedTransports()
     {
-        if ($this->serviceDetails->transportType !== Transport::REST) {
```
Hectorhammett, line 481: How can I run a test that uses the gRpc only?

rules_php_gapic/php_gapic.bzl
bshaffer: I prefer listing the valid transports than just saying it's invalid.

src/Generation/GapicClientV2Generator.php
bshaffer: We would need to update one of the test clients, like we have here:

https://github.com/googleapis/gapic-generator-php/blob/main/tests/Integration/BUILD.bazel#L126

But use a different client (or add a new one)

composer.json
```diff
@@ -26,7 +26,8 @@
             "Testing\\DisableSnippets\\": "tests/Unit/ProtoTests/DisableSnippets/out/src",
             "Testing\\GrpcServiceConfig\\": "tests/Unit/ProtoTests/GrpcServiceConfig/out/src",
             "Testing\\ResourceNames\\": "tests/Unit/ProtoTests/ResourceNames/out/src",
-            "Testing\\RoutingHeaders\\": "tests/Unit/ProtoTests/RoutingHeaders/out/src"
+            "Testing\\RoutingHeaders\\": "tests/Unit/ProtoTests/RoutingHeaders/out/src",
+            "Testing\\BasicGrpcOnly\\": "tests/Unit/ProtoTests/BasicGrpcOnly/out/src"
```
bshaffer, line 30: nit - list these alphabetically

tests/Unit/ProtoTests/BasicGrpcOnly/basic_grpc_only_service.yaml
```diff
@@ -0,0 +1,10 @@
+type: google.api.Service
+config_version: 3
+
+authentication:
+  rules:
+  - selector: 'testing.basicGrpcOnly.BasicGrpcOnly.*'
+    oauth:
+      canonical_scopes: |-
+        scope1,
+        scope2
```
bshaffer, line 10: same here, we may be able to get rid of this file or have it be a lot smaller

tests/Unit/ProtoTests/ClientTest.php
```diff
@@ -35,4 +36,12 @@ public function testUnsupportedTransportThrowsException()
             'credentials' => new InsecureCredentialsWrapper(),
         ]);
     }
+
+    public function testGrpcOnlyClientThrowsExceptionForRestTransport()
+    {
+        $this->expectException(ValidationException::class);
+        $this->expectExceptionMessage('Unexpected transport option "rest". Supported transports: grpc, grpc-fallback');
+
+        $client = new BasicGrpcOnlyClient(['transport' => 'rest']);
```
bshaffer, line 45: I think we can fix the failing tests by passing `Google\ApiCore\InsecureCredentialsWrapper`:

```suggestion
        $client = new BasicGrpcOnlyClient([
            'transport' => 'rest',
            'credentials' => new InsecureCredentialsWrapper(),
        ]);
```

src/Generation/GapicClientV2Generator.php
```diff
@@ -507,29 +507,31 @@ private function construct(): PhpClassMember
...
+            'rest' => AST::arrayEllipsis()
+        ];
 
-        $transportConfigSampleValues = [];
-        if ($isGrpcRest) {
-            $transportConfigSampleValues['grpc'] = AST::arrayEllipsis();
+        if (Transport::isGRPCOnly($transportType)) {
+            unset($transportConfigSampleValues['rest']);
+        }
+        if (Transport::isRestOnly($transportType)) {
```
bshaffer, line 531: mutually exclusive, should be `elseif`

```suggestion
        } elseif (Transport::isRestOnly($transportType)) {
```

src/Generation/GapicClientV2Generator.php
```diff
@@ -507,29 +507,31 @@ private function construct(): PhpClassMember
...
-        // Assumes there are only two transport types.
-        $isGrpcRest = $this->serviceDetails->transportType === Transport::GRPC_REST;
-
-        $restTransportDocText = 'At the moment, supports only `rest`.';
-        $grpcTransportDocText = 'May be either the string `rest` or `grpc`. Defaults to `grpc` if gRPC support is detected on the system.';
         $transportDocText =
             PhpDoc::text(
                 'The transport used for executing network requests. ',
-                $isGrpcRest ? $grpcTransportDocText : $restTransportDocText,
+                $this->transportDocText(),
```
bshaffer, line 513: nit, but pass `$transportType` as a parameter here to make deps more clear

src/Utils/Transport.php
```diff
@@ -44,4 +44,24 @@ public static function parseTransport(?string $transport): int
...
+    {
+        return Transport::compareTransports(Transport::REST, $transport);
+    }
+
+    public static function isGRPCOnly(int $transport): bool
+    {
+        return Transport::compareTransports(Transport::GRPC, $transport);
+    }
+
+    public static function isGRPCRest(int $transport): bool
```
bshaffer, line 58: nit, but convention is to camel-case gRPC as `Grpc` in class and method names

```suggestion
    public static function isGrpcRest(int $transport): bool
```

tests/Integration/goldens/redis/src/V1/Client/CloudRedisClient.php
```diff
@@ -131,7 +131,6 @@ private static function getClientDefaults()
             'apiEndpoint' => self::SERVICE_ADDRESS . ':' . self::DEFAULT_SERVICE_PORT,
             'clientConfig' => __DIR__ . '/../resources/cloud_redis_client_config.json',
             'descriptorsConfigPath' => __DIR__ . '/../resources/cloud_redis_descriptor_config.php',
-            'gcpApiConfigPath' => __DIR__ . '/../resources/cloud_redis_grpc_config.json',
```
bshaffer, line 134: This shouldn't be removed - see this line, which should be changed to be `!== Transport::REST`:

https://github.com/googleapis/gapic-generator-php/blob/46dc31c550399de766d877038ec7d188f13fcfc9/src/Generation/GapicClientGenerator.php#L446

tests/Integration/goldens/redis/src/V1/Gapic/CloudRedisGapicClient.php
```diff
@@ -347,9 +346,8 @@ public function resumeOperation($operationName, $methodName = null)
      *           default this settings points to the default client config file, which is
      *           provided in the resources folder.
      *     @type string|TransportInterface $transport
-     *           The transport used for executing network requests. May be either the string
-     *           `rest` or `grpc`. Defaults to `grpc` if gRPC support is detected on the system.
-     *           *Advanced usage*: Additionally, it is possible to pass in an already
+     *           The transport used for executing network requests. At the moment, supports only
+     *           `rest`. *Advanced usage*: Additionally, it is possible to pass in an already
```
bshaffer, line 350: The deprecated clients are still incorrect - We should either make the _same_ changes to `GapicClientGenerator.php`, or _preferrably_ continue to throw the "grpc is not supported" error if V1 clients are requested (e.g. if `transport` is `grpc` AND `migration_mode` is anything other than `NEW_SURFACE_ONLY`, throw an error). If we do the latter, we'll need to update these integration tests to be `NEW_SURFACE_ONLY` so they don't throw an error (or add a new Integration client).

src/Generation/GapicClientV2Generator.php
Hectorhammett: Changed it.

I just personally do not really like the use of `else` and `elseif`. Is jut a nitpick of mine 😅

tests/Integration/goldens/redis/src/V1/Client/CloudRedisClient.php
Hectorhammett: I made this change and tested it locally and lots of golden removed this line. Shouldn't be "or" `Transport::REST`?

bshaffer: Hmm that's weird, since the logic should be equivalent for all but the new case of gRPC only, which we just added. It should have gone from "when this is gRPC+rest" to "then this ISNT rest only"

Hectorhammett: Ahhh found the issue. It was between the keyboard and the chair.

I forgot to flip the `===` to `!==`. My bad.

tests/Integration/goldens/redis/src/V1/Gapic/CloudRedisGapicClient.php
Hectorhammett: @bshaffer 
I am confused, why is the one on the`Client` client folder flipped to the correct one and not this one?
image

src/Utils/Transport.php
```diff
@@ -38,9 +39,29 @@ public static function parseTransport(?string $transport): int
...
+    {
+        return Transport::compareTransports(Transport::GRPC_REST, $transport);
+    }
+
+    private static function compareTransports(int $transportA, int $transportB): bool
+    {
+        return $transportA === $transportB;
+    }
```
bshaffer, line 66: nit: do we really need this method? I think it's just as readable (and I would say more readable since there's fewer levels of redirection) to just do the comparison in the methods themselves.

tests/Integration/goldens/redis/src/V1/Client/CloudRedisClient.php
bshaffer: I believe we still need to regenerate the Integration (golden) tests and push the result, as this change should be reverted.

tests/Integration/goldens/redis/src/V1/Gapic/CloudRedisGapicClient.php
bshaffer: Because we have both the `GapicClientGenerator` and the `GapicClientV2Generator`. It's admittedly confusing, and will only be there until we release V2 for all our client libraries (which should happen soon).

src/Generation/GapicClientGenerator.php
```diff
@@ -443,7 +443,7 @@ private function getClientDefaults(): PhpClassMember
 
         // TODO: Consolidate setting all the known array values together.
         // We do this here to maintain the existing sensible ordering.
-        if ($this->serviceDetails->transportType === Transport::GRPC_REST) {
+        if ($this->serviceDetails->transportType !== Transport::REST) {
```
bshaffer, line 446: Ahh I see it now... it's because we need to have this same change in `src/Generation/GapicClientV2Generator.php`, and we want the changes below in `src/Generation/GapicClientGenerator.php`. Either that or we need to throw an error for V1 clients. I am going to look into this!

bshaffer: Okay we can fix this by adding the following to `php_gapic.bzl`:

```bzl
if transport == "grpc" and migration_mode != "NEW_SURFACE_ONLY":
    fail("Error: 'grpc' transport is only supported with 'NEW_SURFACE_ONLY' migration mode")
```

bshaffer: See https://github.com/googleapis/gapic-generator-php/pull/708/files - I show how to make this happen

src/Utils/Transport.php
Hectorhammett: Hmmm is not necessary. I prefer to do it this way, for me sounds weird to have to repeat the same code and it has the added benefit of making it public in the future if we need it.

That being said, that opinion is just an opinion and I will not die on this hill for this. I can change it.

bshaffer: I totally agree it's just a matter of opinion, and I would also prefer to die on a different hill.

src/Generation/GapicClientV2Generator.php
```diff
@@ -661,6 +677,19 @@ private function construct(): PhpClassMember
...
+    {
+        if (Transport::isRestOnly($transportType)) {
+            return 'At the moment, supports only `rest`.';
+        }
+
+        if (Transport::isGrpcOnly($transportType)) {
+            return 'At the moment, supports only `grpc`.';
+        }
+
+        return 'May be either the string `rest` or `grpc`. Defaults to `grpc` if gRPC support is detected on the system.';;
```
bshaffer, line 690: remove extra semicolon 
```suggestion
        return 'May be either the string `rest` or `grpc`. Defaults to `grpc` if gRPC support is detected on the system.';
```

tests/Unit/ProtoTests/BasicGrpcOnly/basic-grpc-only.proto
```diff
@@ -0,0 +1,14 @@
...
+
+// php_namespace option not included; to test generating namespace from proto package.
+
+import "google/api/annotations.proto";
+import "google/api/client.proto";
+import "google/api/field_behavior.proto";
+
+// This is a basicGrpcOnly service.
+service BasicGrpcOnly {
+  option (google.api.default_host) = "basicGrpcOnly.example.com";
```
bshaffer, line 13: nit, but api hosts are lowercase
```suggestion
  option (google.api.default_host) = "basicgrpconly.example.com";
```

bshaffer (approved):

bshaffer:
It worked! https://github.com/googleapis/google-cloud-php/pull/7245/commits/c17dce0a31b5e13c4a48ad59205a463c4167bd6b

Hectorhammett:
@bshaffer Let's gooooo!

PR #708, opened 2024-04-25
Suggestions from code review for gRPC-only feature

bshaffer:
Code review for https://github.com/googleapis/gapic-generator-php/pull/707


src/Generation/GapicClientGenerator.php
```diff
@@ -443,7 +443,7 @@ private function getClientDefaults(): PhpClassMember
 
         // TODO: Consolidate setting all the known array values together.
         // We do this here to maintain the existing sensible ordering.
-        if ($this->serviceDetails->transportType !== Transport::REST) {
+        if ($this->serviceDetails->transportType === Transport::GRPC_REST) {
```
bshaffer, line 446: reverting this change simply because `GapicClientGenerator` is considered "deprecated". But this isn't harming anyone how it was.

tests/Integration/BUILD.bazel
```diff
@@ -220,6 +220,7 @@ php_gapic_library(
     srcs = ["@com_google_googleapis//google/cloud/functions/v1:functions_proto_with_info"],
     grpc_service_config = "@com_google_googleapis//google/cloud/functions/v1:functions_grpc_service_config.json",
     migration_mode = "NEW_SURFACE_ONLY",
+    transport = "grpc",
```
bshaffer, line 223: Moved the "grpc-only" client to be Functions instead of Redis, since now it must be a library marked as `NEW_SURFACE_ONLY`

rules_php_gapic/php_gapic.bzl
```diff
@@ -71,12 +71,14 @@ def php_gapic_srcjar(
         transport = "grpc+rest"
     if transport != "grpc+rest" and transport != "rest" and transport != "grpc":
         fail("Error: Only 'grpc+rest', 'rest' or `grpc` transports are supported")
+    if transport == "grpc" and migration_mode != "NEW_SURFACE_ONLY":
+        fail("Error: 'grpc' transport is only supported with 'NEW_SURFACE_ONLY' migration mode")
```
bshaffer, line 75: this ensures that the previous client surfaces (e.g. `GapicClientGenerator`, as opposed to `GapicClientV2Generator`) are not able to be generated with "grpc-only" transport, for simplicity.

PR #713, opened 2024-05-21
feat: use createNewSurfaceOperationsClient for NEW_SURFACE_ONLY clients

bshaffer:
Depends on https://github.com/googleapis/gax-php/pull/571

Uses new `GapicClientTrait::createNewSurfaceOperationsClient` method when clients have `migration_mode = NEW_SURFACE_ONLY`.

In these cases, the clients are either pre-1.0 or already 2.0, so they can safely use the new surface clients.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
-        "google/gax": "^1.19.1"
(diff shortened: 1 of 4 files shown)


bshaffer:
I think this direction is not a good idea, and it inflates an already-inflated `GapicClientTrait`.

Better to generate the logic directly in the GAPIC client.

bshaffer:
closed in favor of https://github.com/googleapis/gapic-generator-php/pull/714

PR #714, opened 2024-05-21
feat: use new surface LROs in new surface clients

bshaffer:
Ensures new surface clients use new surface operations clients for LRO operations by generating a `createOperationsClient` to override the one in GAX.

alternative to #713 and https://github.com/googleapis/gax-php/pull/571

Depends on https://github.com/googleapis/gax-php/pull/569
Builds on top of https://github.com/googleapis/gapic-generator-php/pull/712


composer.json
```diff
@@ -40,7 +41,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
-        "google/gax": "^1.19.1"
+        "google/gax": "dev-support-operations-for-compute-v2 as 1.34"
```
noahdietz, line 43: Just leaving a comment to update this once release is cut.

src/Generation/GapicClientV2Generator.php
```diff
@@ -285,10 +286,12 @@ private function operationMethods(): Vector
         if (!$this->serviceDetails->hasLro && !$this->serviceDetails->hasCustomOp) {
             return Vector::new([]);
         }
-
-        $ctype = $this->serviceDetails->hasCustomOp ?
-            $this->serviceDetails->customOperationServiceClientType :
-            Type::fromName(OperationsClient::class);
+        $ctype = $this->serviceDetails->hasCustomOp
+            ? $this->serviceDetails->customOperationServiceClientType
+            : Type::fromName($this->serviceDetails->migrationMode === MigrationMode::NEW_SURFACE_ONLY
```
noahdietz, line 291: Should we do this for `MIGRATING` too?

src/Generation/GapicClientV2Generator.php
```diff
@@ -368,6 +371,39 @@ private function operationMethods(): Vector
...
+                : $this->ctx->type(Type::fromName(OperationsClient::class));
+            $createOperationsClient = AST::method('createOperationsClient')
+                ->withAccess(Access::PRIVATE)
+                ->withParams(AST::param(ResolvedType::array(), $options))
+                ->withBody(AST::block(
+                    AST::call(AST::method('unset'))(
+                        AST::index($options, 'serviceName'),
+                        AST::index($options, 'clientConfig'),
+                        AST::index($options, 'descriptorsConfigPath'),
+                    ),
```
noahdietz, line 387: So why the `unset`? Maybe include a comment here including the "why".

src/Generation/GapicClientV2Generator.php
bshaffer: My reasoning here was that, since we've already "Gone GA" with the new surfaces, we need to wait until they're behind a major version (if the library itself is already stable) to make this technically breaking change.

If a library is `NEW_SURFACE_ONLY`, it means either that (1) the library is v2, OR that the library NEVER had a V1 surface and is pre-GA (1.0). In both these scenarios, we can push this "breaking change".  However, for libraries that are "Migrating", they could be 1.0+, and therefore would break with this change.

It's not perfect because we do have some pre-GA libraries which are in "MIGRATING" mode that we COULD push this breaking change out to, but won't be able to until we bump to "NEW_SURFACE_ONLY". I don't really have a good suggestion for that.

src/Generation/GapicClientV2Generator.php
bshaffer: Yes I will do that!

To answer your question, it's a carry-over from [the createOperationsClient method](https://github.com/googleapis/gax-php/blob/710e47dddae8dd490cb02e6bb8f4cb7cadc77572/src/GapicClientTrait.php#L377-L381) in GAX which does the same thing. The reason is that those configs are CLIENT specific, so passing them down to the operations client will break it (it'll load the wrong configs/descriptors).

bshaffer: Added the comment to the generated output

noahdietz (approved):

tests/Unit/autoload.php
```diff
@@ -84,3 +84,5 @@ class_alias(FakeMessage::class, $class);
 
 // Use a custom autoloader to produce fake proto message classes for tests.
 spl_autoload_register('protosOnDemandLoader', true);
+
+require_once __DIR__ . '/../../vendor/autoload.php';
```
bshaffer, line 88:
```suggestion
```

composer.json
bshaffer:
```suggestion
        "google/gax": "^1.34"
```

Issue #717, opened 2024-05-23
EmulatorSupportGenerator doesn't work for clients when transportConfig is already passed.


saranshdhingra:
For clients where transportConfig is already passed during GAPIC client instantiation like `Bigtable`, the credentials passed are invalid and therefore emulators don't work with such clients.

PR #719, opened 2024-05-31
fix: Update Emulator configuration function

ajupazhamayil:
Resolves https://github.com/googleapis/gapic-generator-php/issues/717

### DIFF
--- a/tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php
+++ b/tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php
@@ -1042,7 +1042,7 @@ final class DatabaseAdminClient
     }
 
     /** Configure the gapic configuration to use a service emulator. */
-    private function setDefaultEmulatorConfig(): array
+    private function setDefaultEmulatorConfig(array $options): array
     {
         $emulatorHost = getenv('SPANNER_EMULATOR_HOST');
         if (empty($emulatorHost)) {
--- a/tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php
+++ b/tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php
@@ -376,7 +376,7 @@ final class DatabaseAdminClient
      */
     public function __construct(array $options = [])
     {
-        $options = $options + $this->getDefaultEmulatorConfig();
+        $options = $this->setDefaultEmulatorConfig($options);
         $clientOptions = $this->buildClientOptions($options);
         $this->setClientOptions($clientOptions);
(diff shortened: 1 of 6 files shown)


ajupazhamayil:
Investigating why the code changes didn't update the integration tests goldens!!

ajupazhamayil:
Integration test files are manually updated, introduced some errors to see if the cache delete takes effect. Apparently, the integration tests are running irrespective of the code changes!!

PS: Command used to rotate the cache key: `uuidgen | gh secret set CACHE_VERSION -r https://github.com/googleapis/gapic-generator-php`

cc: @bshaffer

tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php
```diff
@@ -1042,28 +1042,21 @@ public function updateDatabaseDdl(UpdateDatabaseDdlRequest $request, array $call
...
-                    'stubOpts' => [
-                        'credentials' => ChannelCredentials::createInsecure(),
-                    ],
-                ],
-            ],
-            'credentials' => new InsecureCredentialsWrapper(),
-        ];
+        $options['apiEndpoint'] = $emulatorHost;
+        $options['transportConfig']['grpc']['stubOpts']['credentials'] = ChannelCredentials::createInsecure();
+        $options['credentials'] = new InsecureCredentialsWrapper();
```
bshaffer, line 1059: This looks great!! The only problem I see is that where before this was providing default values, it's now overwriting any user-supplied values. What if we use `??=` instead?

```suggestion
        $options['apiEndpoint'] ??= $emulatorHost;
        $options['transportConfig']['grpc']['stubOpts']['credentials'] ??= ChannelCredentials::createInsecure();
        $options['credentials'] ??= new InsecureCredentialsWrapper();
```

bshaffer (commented):
This looks great! Only suggestion is we should use the noal-coalescing assignment operator (`??=`) to ensure that these are defaults that can be overwritten by the user.

See https://github.com/googleapis/gapic-generator-php/pull/720 where I've added the operator to the AST

bshaffer (approved):

tests/Integration/goldens/spanner/src/V1/Client/DatabaseAdminClient.php
ajupazhamayil: Once the customer set the emulator environment variable, endpoint SHOULD be replaced by design, else it will by default hit the production product.

bshaffer: I personally don't like that, I think they should just be defaults... I don't ever think an environment variable should REPLACE something in code - they should _always_ function as a default

PR #736, opened 2024-12-03
feat: add compatibility for PHP 8.4

bshaffer:
Ensure `parseName` does not use implicit nullable

### DIFF
--- a/src/Utils/ResolvedType.php
+++ b/src/Utils/ResolvedType.php
@@ -75,8 +75,8 @@ class ResolvedType
         /** @var Type *Readonly* The type of this resolved-type. */
         public ReadOnly Type $type,
         private Closure $fnToCode,
-        private bool $optional = false)
-    {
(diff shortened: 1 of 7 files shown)


src/Utils/ResolvedType.php
```diff
@@ -69,19 +69,16 @@ public static function self(): ResolvedType
      *
      * @param string $typeName The resolved name of the type.
      */
-    public function __construct(Type $type, \Closure $fnToCode)
+    public function __construct(
+        /** @var Type *Readonly* The type of this resolved-type. */
+        public ReadOnly Type $type,
+        private \Closure $fnToCode,
```
Hectorhammett, line 75: Wouldn't it be better to `use Closure` instead of `\Closure`?

bshaffer: agreed. Done!

PR #746, opened 2025-03-03
fix: ensure hash is int to prevent loss of precision error

bshaffer:


dazuma (approved):
Stamp

tests/Unit/Collections/VectorTest.php
```diff
@@ -347,4 +347,17 @@ public function testToArrayAssoc(): void
         $a = $v->toArray(fn ($x) => "key_{$x}", fn ($x) => $x * 10);
         $this->assertEquals(['key_1' => 10, 'key_2' => 20], $a);
     }
+
+    public function testGetHashDoesNotLosePrecision()
+    {
+        // ensure that get hash always returns an int regardless of size
```
bshaffer, line 353:
```suggestion
        // ensure that get hash always returns an int in case of overflow
```

PR #749, opened 2025-03-06

vchudnov-g:
This comes from updating Java to JDK 17 in the Docker container (cl/733907084).


bshaffer (approved):

bshaffer:
I marked this as a `chore` since it doesn't seem like this actually fixes an issues in the current generator. Please correct me if I'm wrong @vchudnov-g

PR #754, opened 2025-05-30

meltsufin:
See: cl/765308392.


bshaffer:
FYI this broke our "bazel example" tests. I believe we didn't catch this because the bazel cache was hit. 

I am working on a fix in #756

PR #757, opened 2025-06-05

viacheslav-rostovtsev:


bshaffer:
I believe this is a duplicate of https://github.com/googleapis/gapic-generator-php/pull/754

PR #759, opened 2025-06-19
Add the PaginationExceptions class to force a paginated field on an RPC

Hectorhammett:
There has been a change on proto fieles where we were relying on having only one `repeated` field to use it as a `paginated` field but that has not been the case before. 

This PR add support for adding a proto type that is not following this convention to the `PaginationExceptions` class and the generator being able to find and utilize said field as the paginated one.

This PR aims to add support to this new process as well as to add suppor for the `'UsableSubnetworksAggregatedList::items`as the correct paginated field. Further additions might happen on a different PR.

b/425742931


src/Generation/MethodDetails.php
```diff
@@ -134,30 +135,47 @@ private static function maybeCreatePaginated(ServiceDetails $svc, MethodDescript
...
-            if (is_null($resourceByNumber)) {
-                $resourceByNumber = $resourceListCandidates->orderBy(fn ($x) => $x[0])->firstOrNull();
+
+        // If we have the type inside the ExplicitPagination class,
+        // use the field stored in the paginations array
+        if (ExplicitPagination::exists($outputMsg->desc->getFullName())) {
+            $resources = $outputMsg->desc
+                ->getFieldByName(
+                    ExplicitPagination::getPagination($outputMsg->getName())
+                );
```
bshaffer, line 148: nit
```suggestion
            $resources = $outputMsg->desc->getFieldByName(
                ExplicitPagination::getPagination($outputMsg->getName())
            );
```

tests/Unit/ProtoTests/BasicExplicitPaginated/out/src/Client/BasicExplicitPaginatedClient.php
```diff
@@ -0,0 +1,192 @@
...
+     *           Retry settings to use for this call. Can be a {@see RetrySettings} object, or an
+     *           associative array of retry settings parameters. See the documentation on
+     *           {@see RetrySettings} for example usage.
+     * }
+     *
+     * @return ExplicitResponse
+     *
+     * @throws ApiException Thrown if the API call fails.
+     */
+    public function methodExplicitPaginated(ExplicitRequest $request, array $callOptions = []): ExplicitResponse
```
bshaffer, line 188: well lookie here! We caught a bug!

This should be returning a `PagedListResponse`, and it isn't!

tests/Unit/ProtoTests/BasicExplicitPaginated/out/src/Gapic/BasicExplicitPaginatedGapicClient.php
```diff
@@ -0,0 +1,216 @@
...
+
+        if (isset($optionalArgs['partOfRequestB'])) {
+            $request->setPartOfRequestB($optionalArgs['partOfRequestB']);
+        }
+
+        if (isset($optionalArgs['partOfRequestC'])) {
+            $request->setPartOfRequestC($optionalArgs['partOfRequestC']);
+        }
+
+        return $this->startCall('MethodExplicitPaginated', ExplicitResponse::class, $optionalArgs, $request)->wait();
```
bshaffer, line 214: This one should also be returning a `PagedListResponse`, so something must be broken in `Utils\ExplicitPagination`

src/Utils/ExplicitPagination.php
```diff
@@ -0,0 +1,57 @@
...
+class ExplicitPagination
+{
+    /**
+     * This is a list of explicit pagination for the pagination heuristics.
+     * If there is a proto that has multiple repeated fields and we want to
+     * manually override which of the repeated fields is the one should be paginated on
+     * should be added to this list
+     */
+    private static array $paginations = [
+        'Google.Cloud.Compute.V1.UsableSubnetworksAggregatedList' => 'items',
```
bshaffer, line 31: I was assuming this would be the namespace, e.g. `Google\Cloud\Compute\V1\UsableSubnetworksAggregatedList`.  However, what you have here also works... as long as it's correct. In this case, it should be lowercase, to [match the package name](https://raw.githubusercontent.com/googleapis/googleapis/refs/heads/master/google/cloud/compute/v1/compute.proto)
```suggestion
        'google.cloud.compute.v1.UsableSubnetworksAggregatedList' => 'items',
```

src/Generation/MethodDetails.php
```diff
@@ -134,30 +135,47 @@ private static function maybeCreatePaginated(ServiceDetails $svc, MethodDescript
...
-            $resourceByNumber = $resourceMapCandidates->orderBy(fn ($x) => $x[0])->firstOrNull();
-            if (is_null($resourceByNumber)) {
-                $resourceByNumber = $resourceListCandidates->orderBy(fn ($x) => $x[0])->firstOrNull();
+
+        // If we have the type inside the ExplicitPagination class,
+        // use the field stored in the paginations array
+        if (ExplicitPagination::exists($outputMsg->desc->getFullName())) {
+            $resources = $outputMsg->desc
+                ->getFieldByName(
+                    ExplicitPagination::getPagination($outputMsg->getName())
```
bshaffer, line 147: I believe the issue is that this needs to match the `exists` above
```suggestion
                    ExplicitPagination::getPagination($outputMsg->desc->getFullName())
```

googleapis
bshaffer, line 1: Did you mean to update the `googleapis` submodule? It shouldn't be an issue as long as the tests pass, but it's probably unnecessary for this PR

Edit: Looks like this points to a [recent commit from yesterday](https://github.com/googleapis/googleapis/commit/d931df43c4686be708ddfee74ee3be64571af0a2), so we might as well keep it

bshaffer (dismissed):

bshaffer (approved):

bshaffer (dismissed):

PR #765, opened 2025-07-14
feat: Add exception rules for multiple protos on the compute clients

Hectorhammett:
Edit the generation logic to pick the correct paginated results from a proto message.

b/427528266


bshaffer (commented):
So we don't want to make exceptions for these - the only reason we did that for `UsableSubnetworksAggregatedList` was because 1) it was a breaking change and we wanted it to unblock the release, and 2) it doesn't seem to obey any heuristic, so we needed an exception.

But implementing a better heuristic for determining which RPCs are paginated will result in a breaking change, as we are going to enforce a _new_ heuristic which will correct the existing RPCs which we previously missed, as well as ensure we don't miss turning RPCs into paginated lists in the future.

The PHP Heuristic should look like this:

 - Check the exception dictionary. If the method is there as a key, use the value as the results field (you've done this already)
 - If there is exactly one map field, that field is the results field.
 - If there are no repeated fields there is no results field and we shall throw an error
 - If there is exactly one repeated field, that field is the results field.
 - If there is one repeated message field and any number of repeated primitives, we page on the message field

bshaffer (commented):
I added a few minor suggestions, but the logic here looks good! We just need to add some tests (either by updating `compute_small.proto` or by creating a new test like `rest_only_pagination.proto`.

PR #774, opened 2025-08-24
Remove final from client without interfaces

VincentLanglet:
Resolve https://github.com/googleapis/google-cloud-php/pull/8492

Currently:
- It's not possible to mock any ServiceClient like the TranslateServiceClient because
  - It's final
  - It does not implements an interface

- The proposed solution of using bypass-final has drawbacks
  - the lib bypass-final is highly opinionated and source of debate in php community
  - it does not resolve all issues because other tools like static analysis won't understand the fact a final class is mocked/extended.

- Also, `*ServiceClient` classes are the ONLY ones final in the huge google codebase, and this doesn't seems needed/have benefit.

Therefor, I tried to remove the final keyword in the generation.

### DIFF
--- a/src/Generation/GapicClientV2Generator.php
+++ b/src/Generation/GapicClientV2Generator.php
@@ -108,7 +108,7 @@ class GapicClientV2Generator
     {
         return AST::class(
                 $this->serviceDetails->gapicClientV2Type,
-                final: true)
+                )
             ->withPhpDoc(PhpDoc::block(
                 PhpDoc::preFormattedText(
                     $this->serviceDetails->docLines->skip(1)
--- a/tests/Integration/goldens/functions/src/V1/Client/CloudFunctionsServiceClient.php
+++ b/tests/Integration/goldens/functions/src/V1/Client/CloudFunctionsServiceClient.php
@@ -79,7 +79,7 @@ use Psr\Log\LoggerInterface;
  * @method PromiseInterface<TestIamPermissionsResponse> testIamPermissionsAsync(TestIamPermissionsRequest $request, array $optionalArgs = [])
  * @method PromiseInterface<OperationResponse> updateFunctionAsync(UpdateFunctionRequest $request, array $optionalArgs = [])
  */
-final class CloudFunctionsServiceClient
+class CloudFunctionsServiceClient
 {
     use GapicClientTrait;
     use ResourceHelperTrait;
--- a/tests/Integration/goldens/redis/src/V1/Client/CloudRedisClient.php
+++ b/tests/Integration/goldens/redis/src/V1/Client/CloudRedisClient.php
@@ -94,7 +94,7 @@ use Psr\Log\LoggerInterface;
  * @method PromiseInterface<Location> getLocationAsync(GetLocationRequest $request, array $optionalArgs = [])
  * @method PromiseInterface<PagedListResponse> listLocationsAsync(ListLocationsRequest $request, array $optionalArgs = [])
  */
-final class CloudRedisClient
+class CloudRedisClient
 {
     use GapicClientTrait;
     use ResourceHelperTrait;
--- a/tests/Integration/goldens/securitycenter/src/V1/Client/SecurityCenterClient.php
+++ b/tests/Integration/goldens/securitycenter/src/V1/Client/SecurityCenterClient.php
@@ -154,7 +154,7 @@ use Psr\Log\LoggerInterface;
  * @method PromiseInterface<SecurityMarks> updateSecurityMarksAsync(UpdateSecurityMarksRequest $request, array $optionalArgs = [])
  * @method PromiseInterface<Source> updateSourceAsync(UpdateSourceRequest $request, array $optionalArgs = [])
  */
-final class SecurityCenterClient
+class SecurityCenterClient
 {
     use GapicClientTrait;
     use ResourceHelperTrait;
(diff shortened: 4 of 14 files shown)


bshaffer:
@VincentLanglet Thanks for submitting this. We appreciate your feedback and your suggestions!

> the lib bypass-final is highly opinionated and source of debate in php community

Can you elaborate on this?

> it does not resolve all issues because other tools like static analysis won't understand the fact a final class is mocked/extended.

A few. things here - because mocking takes place in a test suite, I do not find this to be a compelling reason. I am not going to tell people to NOT use static analysis tools on their test suites, but I think it's an edge-case. Not to mention, static analyzers can easily be configured to bypass errors like this.

> this doesn't seems needed/have benefit.

The benefit of marking a class as final is that you guarantee that the class will be used as you intend it, and not be modified / extended in unexpected ways. This is something that has happened to us with our client libraries in the past - we had bugs filed when things broke because certain methods were being extended which shouldn't be.

That being said, I do sympathize with your situation, and it's strange to not be able to mock a client without pulling in a 3rd party library. Maybe there are other things we can do to prevent the clients from being used in unexpected ways, such as ensuring there are no `protected` methods, and using `self` instead of `static`. But this will take some time.

I hope for the time being, using [dg/bypass-finals](https://github.com/dg/bypass-finals) is an appropriate workaround.

VincentLanglet:
> @VincentLanglet Thanks for submitting this. We appreciate your feedback and your suggestions!
> [...]
> That being said, I do sympathize with your situation, and it's strange to not be able to mock a client without pulling in a 3rd party library.

Thanks to you @bshaffer for taking time to find a solution.

> > the lib bypass-final is highly opinionated and source of debate in php community
> 
> Can you elaborate on this?

I cannot find the discussion I saw on twitter (I think 1-2 years ago ?) but this was somehow heated. Maybe because mains actors were kinda Laravel VS Symfony. The Laravel philosophy is more magic and letting the user doing everything he wants, while the Symfony one is more rigourous/strict (based on the point of view).

Main argument in favor of not using bypass-final is the fact it's a code-smell:
- If you need is reasonable, maybe there is another way to do it or maybe there should be another way to do it (and the lib should be improved)
- If the author of a lib is marking a class final, it's not to allow someone to remove the final keyword and still extending it. You're saying:
"The benefit of marking a class as final is that you guarantee that the class will be used as you intend it, and not be modified / extended in unexpected ways. This is something that has happened to us with our client libraries in the past - we had bugs filed when things broke because certain methods were being extended which shouldn't be."

And this is technically still possible because someone could use bypass-final to remove the final keyword and still extending your client. That's why bypass-final is not liked by everyone.

One could also argue that if the maintainers of a lib has to themself use bypass-final to test the lib, it might means that there is something to change/improve too.

> > it does not resolve all issues because other tools like static analysis won't understand the fact a final class is mocked/extended.
> 
> A few. things here - because mocking takes place in a test suite, I do not find this to be a compelling reason. I am not going to tell people to NOT use static analysis tools on their test suites, but I think it's an edge-case. Not to mention, static analyzers can easily be configured to bypass errors like this.
>
> I hope for the time being, using [dg/bypass-finals](https://github.com/dg/bypass-finals?rgh-link-date=2025-08-26T16%3A41%3A48Z) is an appropriate workaround.

While I agree that
- Static analysis on test folder might be considered as a edge-case
- Lot of static analysis errors can be ignored (for instance with a baseline for PHPStan)

Unfortunately in the case of [extending/mocking final class](https://github.com/phpstan/phpstan/issues/8901), the issue are not ignorable by PHPStan because it consider it's kinda a fatal error which should crash php (and therefor shouldn't be ignorable).

So currently bypass-finals is not an appropriate workaround for me.

> The benefit of marking a class as final is that you guarantee that the class will be used as you intend it, and not be modified / extended in unexpected ways. This is something that has happened to us with our client libraries in the past - we had bugs filed when things broke because certain methods were being extended which shouldn't be.
>
> Maybe there are other things we can do to prevent the clients from being used in unexpected ways, such as ensuring there are no `protected` methods, and using `self` instead of `static`. But this will take some time.

What I first don't understand is the fact that only few classes are final then.
For instance, for google-cloud-php/Translate, unfortunately the ONLY class I need to mock is the final one
https://github.com/googleapis/google-cloud-php/blob/7de4a722e38ad67c727d7b5610c4952fb5004240/Translate/src/V3/Client/TranslationServiceClient.php#L143

But 
- all the other class of `google-cloud-php/Translate` are not final
- the `Google\LongRunning\Client\OperationsClient` (returned by getOperations) which is also a client (and use the GapicClientTrait) is not final.

Anyway, I agree final has benefits and saw generally this pattern with the following implementation
- Class without too much/complicated logic/the need of external services are considered as DTO/ValueObject and are final without extra needs.
- Class with complex logic/need of external services are final but implements an interface.

As a short solution you might accepts to consider would be to change `final class Foo` into
```
/** 
 * @final 
 */
class Foo
```
This would basically tell the developer he should not extends the class (and therefor you don't provide BC on it) but, you still allow mocks.

The best solution IMHO would be to introduce interfaces for the client ([as recommended by the PHPStan creator](https://x.com/OndrejMirtes/status/1622311556531867651)). You would have
```
interface TranslationClientInterface
{
    public function translateText(TranslateTextRequest $request, array $callOptions = []): TranslateTextResponse;
    
    // all the other public methods
}

final class TranslationServiceClient implement TranslationClientInterface
```

So the developer in his code
- use TranslationClientInterface to type param/return type of his function/methods
- use the TranslationServiceClient when he needs a real implementation
- mocks the TranslationClientInterface in phpunit

Since you use a "php-generator" to generate classes, how long/hard do you think it would be @bshaffer to generate an interface for the clients when the class is generated ?

PR #776, opened 2025-08-26
feat: add "database" request param for Firestore::listen

bshaffer:
Addresses https://github.com/googleapis/google-cloud-php/pull/8267

TODO:
 - [X] ~~Look at how to do this with the AST, instead of having PHP code in strings (which is scary)~~ moved to heredoc instead - since the code is static there's no need to use an AST
 - [x] Pull out duplicate methods into a `PostProcessorTrait`


bshaffer:
This PR is an abomination and its author should be scorned from this repo forevermore

PR #785, opened 2025-10-23
fix: prevent reserved variable names in tests

bshaffer:
See https://github.com/googleapis/google-cloud-php/pull/8683

[Tests are failing](https://github.com/googleapis/google-cloud-php/actions/runs/18725681941/job/53409593436?pr=8683#step:8:338) with the following error: 
Screenshot 2025-10-23 at 3 29 10 PM

This is because [in one of the generated tests](https://github.com/googleapis/google-cloud-php/pull/8683/files#diff-5056250686514826953dd103216e679fbb21f07faa9a2e0b09bc17cb81bbd338R378-R414), the `$status` local variable in the test is being overwritten with a variable named `$status` that is used to populate a required field in the request:

```php
    public function searchAdReviewCenterAdsExceptionTest()
    {
        //...
        $status = new stdClass();
        $status->code = Code::DATA_LOSS;
        $status->details = 'internal error';
        // ...
        $status = AdReviewCenterAdStatus::AD_REVIEW_CENTER_AD_STATUS_UNSPECIFIED;
        $request = (new SearchAdReviewCenterAdsRequest())->setParent($formattedParent)->setStatus($status);
        try {
            $gapicClient->searchAdReviewCenterAds($request);
            // If the $gapicClient method call did not throw, fail the test
            $this->fail('Expected an ApiException, but no exception was thrown.');
        } catch (ApiException $ex) {
            $this->assertEquals($status->code, $ex->getCode());
            $this->assertEquals($expectedExceptionMessage, $ex->getMessage());
        }
        //...
    }
```

I've fixed this by passing an array of `$reservedNames` to the code which generates this, and prefixing those names with `request_` before they're converted to camelcased. This will result in the name being `$requestStatus` in this case.

### DIFF
--- a/tests/Unit/ProtoTests/Basic/out/tests/Unit/Client/BasicClientTest.php
+++ b/tests/Unit/ProtoTests/Basic/out/tests/Unit/Client/BasicClientTest.php
@@ -129,9 +129,11 @@ class BasicClientTest extends GeneratedTest
         // Mock request
         $aString = 'aString-929604177';
         $partOfRequestA = [];
+        $status = 892481550;
         $request = (new RequestWithArgs())
             ->setAString($aString)
-            ->setPartOfRequestA($partOfRequestA);
+            ->setPartOfRequestA($partOfRequestA)
+            ->setStatus($status);
         $response = $gapicClient->methodWithArgs($request);
         $this->assertEquals($expectedResponse, $response);
         $actualRequests = $transport->popReceivedCalls();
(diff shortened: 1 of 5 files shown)


tests/Unit/ProtoTests/Basic/out/tests/Unit/Client/BasicClientTest.php
```diff
@@ -167,9 +171,11 @@ public function methodWithArgsExceptionTest()
         // Mock request
         $aString = 'aString-929604177';
         $partOfRequestA = [];
+        $requestStatus = 892481550;
```
bshaffer: it worked!

Hectorhammett (approved):

Issue #813, opened 2026-04-10
Feature Request: Implement interfaces for GAPIC service classes


bshaffer:
To resolve issues with the `final` keyword (see https://github.com/googleapis/google-cloud-php/issues/8588), we can implement and release interfaces for each service class.

We should consider implementing #783 and creating a root `Google/ApiCore/ServiceInterface` as part of this effort

PR #818, opened 2026-04-29
feat: add interfaces for service clients

bshaffer:
addresses https://github.com/googleapis/gapic-generator-php/issues/813


bshaffer:
Failing "Bazel Build" tests are not caused by these changes

Issue #195, opened 2021-03-10
Clean up all TODO(vNext) technical debt


miraleung:

bshaffer:
There are 57 instances of `@TODO(vNext)` in the generator - all of which I believe are very minor - they're mostly consistency changes that complicate the code but would be a bandaid rip WRT reviewing the PR changes. For example, an instance where a comment exists `// Mock Request` in _some_ unit tests but not all of them, simply because that's how it was in the previous generator and we didn't want to cause too much churn.

Bumping this up in priority

PR #834, opened 2026-06-23
chore: remove V1 surface

bshaffer:
- removed `migration_mode` support - this bazel option is now ignored, and all generated code is V2
- removed V1 samples gen, v1 client gen, and v1 unit test gen. Removed "V2" from the classname of all V2 classes (as it's now the only version)
- removed V1-specific structures such as GapicEnum and OneOf wrapping classes
- updated all unit and integration tests


rules_php_gapic/php_gapic.bzl
```diff
@@ -54,10 +54,13 @@ def php_gapic_srcjar(
         transport = None,
         rest_numeric_enums = False,
         generate_snippets = True,
-        # Supported values validated and specified in src/Utils/MigrationMode.php.
+        # TODO(vNext): Remove deprecated migration_mode parameter once all downstream repos are cleaned up.
```
Hectorhammett, line 57: Let's create a ticket to keep track of this.

Simple but useful 👍

src/Generation/GapicClientGenerator.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2020 Google LLC
+ * Copyright 2022 Google LLC
```
Hectorhammett: Image

Hectorhammett: I mean this was created not this year, but then why the update?

src/Generation/GapicClientGenerator.php
```diff
@@ -34,43 +30,48 @@
...
+
+    public static function generate(SourceFileContext $ctx, ServiceDetails $serviceDetails, bool $generateSnippets): PhpFile
     {
-        return (new GapicClientGenerator($ctx, $serviceDetails))->generateImpl();
+        return (new GapicClientGenerator($ctx, $serviceDetails, $generateSnippets))->generateImpl();
     }
 
     private SourceFileContext $ctx;
     private ServiceDetails $serviceDetails;
+    // TODO(v2): This can be cleaned up after the v2 migration is complete.
```
Hectorhammett: With the recommendation of creating a ticket, let's use that same ticket to keep this tracked.

src/Generation/GapicClientGenerator.php
```diff
@@ -34,43 +30,48 @@
...
         $this->ctx = $ctx;
         $this->serviceDetails = $serviceDetails;
+        $this->generateSnippets = $generateSnippets;
     }
 
     private function generateImpl(): PhpFile
     {
         // TODO(vNext): Remove the forced addition of these `use` clauses.
         $this->ctx->type(Type::fromName(\Google\ApiCore\PathTemplate::class));
+        $this->ctx->type(Type::fromName(\Google\ApiCore\Options\ClientOptions::class));
```
Hectorhammett: nit:
I don't really love using `\` instead of the `use` keyword at the top.

src/Generation/GapicClientGenerator.php
```diff
@@ -92,68 +93,93 @@ private function generateImpl(): PhpFile
...
 
     private function serviceName(): PhpClassMember
     {
         return AST::constant('SERVICE_NAME')
             ->withPhpDocText('The name of the service.')
+            ->withAccess(Access::PRIVATE)
             ->withValue($this->serviceDetails->serviceName);
     }
 
+    private function apiVersion()
```
Hectorhammett: Hmmm I see that we can either return an `AST::property` or just return, can we `return null` on the first case so we can declare the return type?

```php
private function apiVersion(): null|PhpProperty
```

src/Generation/SnippetDetails.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2022 Google LLC
+ * Copyright 2023 Google LLC
```
Hectorhammett: Random year change.

src/Generation/SnippetDetails.php
```diff
@@ -107,57 +127,15 @@ private function handleField(FieldDetails $field, ?string $parentFieldName = nul
...
-        $requestVar = AST::var('request');
-        $this->sampleAssignments = $this->sampleAssignments->append(
-            AST::assign(
-                $requestVar,
-                $value
-            )
-        );
-        $this->rpcArguments = $this->rpcArguments->append([$requestVar]);
-    }
-
```
Hectorhammett, line 160: Nice chunk removal 🔥

src/Generation/UnitTestsGenerator.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2020 Google LLC
+ * Copyright 2022 Google LLC
```
Hectorhammett: Another year change

src/Generation/GapicClientGenerator.php
bshaffer: Because I renamed the files from GapicClientV2Generator, UnitTestV2Generator, and SnippetDetailsV2, the copyright year in THOSE files was copied over.

src/Generation/GapicClientGenerator.php
bshaffer: Actually, this was here from before (it was in the file when I copied it over), and we can just do the cleanup now and remove it

rules_php_gapic/php_gapic.bzl
bshaffer: This is being tracked in https://github.com/googleapis/gapic-generator-php/issues/195

Hectorhammett (approved):

PR #838, opened 2026-07-16
feat: add support for method signatures with a dotted notation.

Hectorhammett:
Using dotted notations on method signatures is valid for a proto file but our generator does not take into account this. This aims to fix this.

### DIFF
--- a/src/Generation/BuildMethodFragmentGenerator.php
+++ b/src/Generation/BuildMethodFragmentGenerator.php
@@ -128,11 +128,12 @@ class BuildMethodFragmentGenerator
         };
 
         $methodSignatureArguments = array_map('trim', array_filter(explode(',', $methodSignature)));
+        $topLevelArguments = array_unique(array_map(fn ($arg) => explode('.', $arg)[0], $methodSignatureArguments));
         $requiredFields = $methodDetails->allFields
-            ->filter(fn ($f) => in_array($f->name, $methodSignatureArguments))
-            ->orderBy(fn ($f) => array_search($f->name, $methodSignatureArguments));
+            ->filter(fn ($f) => in_array($f->name, $topLevelArguments))
+            ->orderBy(fn ($f) => array_search($f->name, $topLevelArguments));
 
-        if (count($methodSignatureArguments) !== $requiredFields->count()) {
+        if (count($topLevelArguments) !== $requiredFields->count()) {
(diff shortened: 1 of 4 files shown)


src/Generation/BuildMethodFragmentGenerator.php
```diff
@@ -128,11 +128,12 @@ public function buildMethodSnippet(MethodDetails $methodDetails, string $methodS
...
         $methodSignatureArguments = array_map('trim', array_filter(explode(',', $methodSignature)));
+        $topLevelArguments = array_unique(array_map(fn ($arg) => explode('.', $arg)[0], $methodSignatureArguments));
         $requiredFields = $methodDetails->allFields
-            ->filter(fn ($f) => in_array($f->name, $methodSignatureArguments))
-            ->orderBy(fn ($f) => array_search($f->name, $methodSignatureArguments));
+            ->filter(fn ($f) => in_array($f->name, $topLevelArguments))
+            ->orderBy(fn ($f) => array_search($f->name, $topLevelArguments));
 
-        if (count($methodSignatureArguments) !== $requiredFields->count()) {
+        if (count($topLevelArguments) !== $requiredFields->count()) {
```
bshaffer, line 136: Are we sure about this change? I feel like we want to keep this as `$methodSignatureArguments`. If you had two required fields `nested.field` and `nested.field2`, `$methodSignatureArguments` would be 2 (and not throw an exception) whereas `$topLevelArguments` would be 1 (and would throw an exception where it shouldn't)

bshaffer (commented):
One small nit, otherwise this looks fantastic!

src/Generation/BuildMethodFragmentGenerator.php
Hectorhammett: The thing is, `$requiredField->count()` returns only the top level fields, not the nested fields, so if we use `$methodSignatureArguments` for the check using your example `['nested.field', 'nested.field2']`:
```php
// $methodSeignatureArguments would be 2: [nested.field, nested.field2] 
// $requiredFields would be 1: [nested]
count($methodSignatureArguments) !== $requiredFields->count()
```

Then it would throw the exception you are mentioning. It seems to be related to the fact that the `$requiredFields` are only being setup as the direct children of a request object, but does not include the nested children of those.

bshaffer (approved):

PR #842, opened 2026-07-20
fix: Avoid passing null as an array offset for longRunning descriptors in resumeOperation

cy-yun:
**Title:** fix: Avoid passing null as an array offset for longRunning descriptors in resumeOperation

**Description:**
This PR fixes a PHP deprecation warning generated by the client libraries when calling `resumeOperation()`.

Previously, `resumeOperation($operationName, $methodName = null)` passed `$methodName` directly as an array offset when checking for `longRunning` descriptors, triggering a deprecation warning if `$methodName` was `null`. The code generator has been updated to safely use `isset($methodName)` before attempting to access the array key.

This improves the Spanner System Test by eliminating deprecation message spamming in the test outputs.

Fixes https://github.com/googleapis/google-cloud-php/issues/9230

### DIFF
--- a/src/Generation/GapicClientGenerator.php
+++ b/src/Generation/GapicClientGenerator.php
@@ -394,8 +394,12 @@ class GapicClientGenerator
             ->withParams(AST::param(null, $operationName), AST::param(null, $methodName, AST::NULL))
             ->withBody(AST::block(
                 AST::assign($options, AST::ternary(
-                    AST::call(AST::ISSET)(AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning']),
-                    AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning'],
+                    AST::call(AST::ISSET)($methodName),
+                    AST::ternary(
+                        AST::call(AST::ISSET)(AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning']),
+                        AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning'],
+                        $default
+                    ),
                     $default
                 )),
                 AST::assign($operation, AST::new($this->ctx->type(Type::fromName(OperationResponse::class)))(
--- a/src/Generation/GapicClientGenerator.php
+++ b/src/Generation/GapicClientGenerator.php
@@ -393,8 +393,9 @@ class GapicClientGenerator
             ->withAccess(Access::PUBLIC)
             ->withParams(AST::param(null, $operationName), AST::param(null, $methodName, AST::NULL))
             ->withBody(AST::block(
-                AST::assign($options, AST::nullCoalescing(
-                    AST::access(AST::THIS, AST::property('descriptors'))[AST::nullCoalescing($methodName, AST::literal("''"))]['longRunning'],
+                AST::assign($options, AST::ternary(
+                    AST::call(AST::ISSET)(AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning']),
+                    AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning'],
                     $default
                 )),
                 AST::assign($operation, AST::new($this->ctx->type(Type::fromName(OperationResponse::class)))(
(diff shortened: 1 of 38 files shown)


bshaffer (changes requested):
We are dropping PHP 8.1 support soon (PHP 8.1 [ended support](https://www.php.net/supported-versions.php) in 2023, and has been EOL since 2025), and this does not affect PHP 8.2 and beyond, so I don't think we need this.

Additionally, I tested this in PHP 8.1 and did not see a deprecation warning, so I am confused by this change:

```php
$a = []; $b = null;
var_dump(isset($a[$b]['c']));
// false
var_dump($a[$b]['c'] ?? 'd');
// 'd'
```

IF we are going to change this line, we should change it to

```diff
- isset($this->descriptors[$methodName]['longRunning']) ? $this->descriptors[$methodName]['longRunning'] : [];
+ $this->descriptors[$methodName]['longRunning'] ?? [];
```

bshaffer (commented):
this is actually happening in PHP 8.5 and needs to be addressed

tests/Integration/goldens/functions/src/V1/Client/CloudFunctionsServiceClient.php
```diff
@@ -153,7 +153,7 @@ public function getOperationsClient()
      */
     public function resumeOperation($operationName, $methodName = null)
     {
-        $options = $this->descriptors[$methodName]['longRunning'] ?? [];
+        $options = isset($methodName) ? $this->descriptors[$methodName]['longRunning'] ?? [] : [];
```
bshaffer: There are the best solutions I came up with for this
```suggestion
        $options = $this->descriptors[$methodName ?? '']['longRunning']) ?? [];
```

or potentially this 
```suggestion
        $options = $methodName && isset($this->descriptors[$methodName]['longRunning'])
            ? $this->descriptors[$methodName]['longRunning']
            : []
```

bshaffer (changes requested):
Reminder: any clients in the `Gapic` directory are irrelevant, as the V1 surface is deprecated and will be removed in #834.

That being said, I think the fix you have is fine, but we should remove the V1 surface changes or wait for https://github.com/googleapis/gapic-generator-php/pull/834 to be merged.

We could ALSO fix this two other ways:

1. More concise but slightly hacky:
    ```php
    $options = $this->descriptors[$methodName ?? '']['longRunning'] ?? [];
    ```

2. More "proper", but more verbose
    ```php
    $options = $methodName && isset($this->descriptors[$methodName]['longRunning'])
        ? $this->descriptors[$methodName]['longRunning']
        : [];
    ```

I am leaning towards the second one, and since this is only for `resumeOperation`, it seems okay to be more verbose (it won't cause too much churn).

A third option is to implement a `LongRunningOperationClientTrait` (which implements the `resumeOperation` method) and add THAT to each client instead... this method is always the same, so there's really no reason for us to write it every time.

src/Generation/GapicClientGenerator.php
```diff
@@ -390,9 +390,8 @@ private function operationMethods(): Vector
             ->withAccess(Access::PUBLIC)
             ->withParams(AST::param(null, $operationName), AST::param(null, $methodName, AST::NULL))
             ->withBody(AST::block(
-                AST::assign($options, AST::ternary(
-                    AST::call(AST::ISSET)(AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning']),
-                    AST::access(AST::THIS, AST::property('descriptors'))[$methodName]['longRunning'],
+                AST::assign($options, AST::nullCoalescing(
+                    AST::access(AST::THIS, AST::property('descriptors'))[AST::nullCoalescing($methodName, AST::literal("''"))]['longRunning'],
```
bshaffer: we do not want to make any changes to the V1 surface (e.g. anything in the namespace `\Gapic\`) because this class has been removed in https://github.com/googleapis/gapic-generator-php/pull/834 and V1 is deprecated.

src/Generation/UnitTestsGenerator.php
```diff
@@ -114,6 +114,18 @@ private function generateImpl(): PhpFile
...
+            $testMethods = $testMethods->append(
+                AST::method('testClientTestCase')
+                    ->withAccess(Access::PUBLIC)
+                    ->withBody(AST::block(
+                        // Add a dummy assertion to silence PHPUnit warning
+                        ($this->assertTrue)(AST::literal('true'))
+                    ))
+            );
+        }
+
```
bshaffer: Same here, this class generates tests for the V1 surface, and it will be removed in https://github.com/googleapis/gapic-generator-php/pull/834

bshaffer (commented):
Everything looks great but we should revert changes to the V1 surface

bshaffer (approved):

PR #847, opened 2026-07-22
fix: Fix the parent resource lookup for more than two levels deep resources

### LINKED ISSUE #846: Unhandled map lookup exception for child resource patterns without matching parent resources.
#### Summary
When generating a PHP GAPIC client from proto files where a message defines child resource references or multi-pattern resource paths (such as `Blurb` in `google/showcase/v1beta1/messaging.proto`), `ProtoCatalog::__construct()` crashes with an unhandled map lookup exception:
```text
In Map.php line 158:
Key users/{user}/profile/blurbs does not exist, called from ProtoCatalog::__construct():163:175
```
---
#### Root Cause
In `src/Utils/ProtoCatalog.php` (lines 171–175), `ProtoCatalog` calculates parent patterns for child resources by stripping the last two path segments (`parentPattern()`):
```php
$parentPatterns = $childPatterns->map(fn ($childPattern) => static::parentPattern($childPattern));
return [$childType, $parentPatterns->map(fn ($x) => $this->resourcesByPattern[$x])->orderBy(fn ($x) => $order[$x->getType()])];
```
This logic makes two unsafe assumptions:
1. **Derived parent paths are assumed to always be registered resources**: For multi-segment or nested subpaths (e.g. `users/{user}/profile/blurbs/{blurb}` or legacy patterns like `.../blurbs/legacy/{legacy_user}~{blurb}`), stripping the last 2 segments yields `users/{user}/profile` or `users/{user}/profile/blurbs`, which are not standalone registered resource patterns in the catalog (e.g. `User` is registered as `users/{user}`).
2. **Unchecked Map indexing**: `$this->resourcesByPattern[$x]` assumes the key `$x` is guaranteed to exist in the catalog map. When `$x` is not in `$this->resourcesByPattern`, the immutable `Map` data structure throws an unhandled exception instead of skipping the non-existent parent pattern.

Hectorhammett:
Closes: #846


src/Utils/ProtoCatalog.php
```diff
@@ -172,7 +172,7 @@ public function __construct(Vector $fileDescs)
                 // Ordering matters, to exactly reproduce monolith behaviour.
                 // parents must be in file order of option-defined resource, followed by file order of message-defined resource.
                 // This is important when selecting a resource to use in test and example methods.
-                return [$childType, $parentPatterns->map(fn ($x) => $this->resourcesByPattern[$x])->orderBy(fn ($x) => $order[$x->getType()])];
+                return [$childType, $parentPatterns->filter(fn ($x) => isset($this->resourcesByPattern[$x]))->map(fn ($x) => $this->resourcesByPattern[$x])->orderBy(fn ($x) => $order[$x->getType()])];
```
bshaffer: This line is over 200 characters long, can we format it better?
```suggestion
                return [
                    $childType,
                    $parentPatterns
                        ->filter(fn ($x) => isset($this->resourcesByPattern[$x]))
                        ->map(fn ($x) => $this->resourcesByPattern[$x])
                        ->orderBy(fn ($x) => $order[$x->getType()])
                ];
```

bshaffer (commented):
This is LGTM except for the formatting change

src/Utils/ProtoCatalog.php
Hectorhammett: Oh woops yeah, let me fix this.

bshaffer (approved):

PR #854, opened 2026-08-09
fix: remove empty fragments directories from packaged output

### LINKED ISSUE #621: V2 post-processing step leaves "fragments" directories
This is of course very low priority, as it does not cause any issues AFAIK, but when untar-ing the resulting package from the GAPIC post-processor, I can see that the `fragment` directories are there even though they contain no files (and we don't want them there):

```
$ bazel build google/cloud/support/v2:google-cloud-support-v2-php
$ tar -xvf bazel-bin/google/cloud/support/v2/google-cloud-support-v2-php.tar.gz
google-cloud-support-v2-php/fragments/
google-cloud-support-v2-php/fragments/Google/
google-cloud-support-v2-php/fragments/Google/Cloud/
google-cloud-support-v2-php/fragments/Google/Cloud/Support/
google-cloud-support-v2-php/fragments/Google/Cloud/Support/V2/
google-cloud-support-v2-php/proto/
...
```
We should remove the empty directories in the post processing step to reduce the noise

amarkdotdev:
## Summary
- Delete `fragments/` after fragment injection in the post-processor.
- Also prune any leftover `fragments/` tree in `php_gapic_pkg.bzl` before creating the package tarball.
- Update the fragment injection unit test to assert the directory is removed.

Fixes #621

## Notes
- Contributor will sign the Google CLA if required for this repository.

## Test plan
- [ ] Run `FragmentInjectionProcessorTest`
- [ ] Build a PHP GAPIC package and confirm the tarball has no empty `fragments/` directories

### DIFF
--- a/rules_php_gapic/php_gapic_pkg.bzl
+++ b/rules_php_gapic/php_gapic_pkg.bzl
@@ -70,6 +70,8 @@ def _php_gapic_src_pkg_impl(ctx):
         fi
     done
     {post_processor} --input {package_dir_path}
+    # Drop any leftover empty fragments/ tree from the package tarball.
+    rm -rf {package_dir_path}/fragments
     cd {package_dir_path}/{tar_cd_suffix}
     tar --exclude=*.build.txt  -zchpf {tar_prefix}/{package_dir}.tar.gz {tar_prefix}/*
     cd -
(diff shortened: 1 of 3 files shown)


amarkdotdev:
rebased onto main

PR #855, opened 2026-08-12
fix: gracefully handle missing proto/src directories in ProtobufNoCheckInHeaderProcessor

### LINKED ISSUE #7260: Fatal Error in PHP Post-Processor when generating AccessApproval
The gapic-generator-php post-processor crashes with a fatal UnexpectedValueException when generating libraries. 

Conclusion: The stack trace pointed to RecursiveDirectoryIterator::__construct(././proto/src). Inspecting ProtobufNoCheckInHeaderProcessor.php revealed it unconditionally tries to iterate over a proto/src directory. However, components in google-cloud-php (like AccessApproval) often lack this folder, causing the iterator to crash instantly. Fix: Add an is_dir() check before instantiating the iterator to ensure the script safely skips components that don't have a proto/src directory.

sofisl:
This PR adds a safety check to ProtobufNoCheckInHeaderProcessor to ensure the proto/src directory actually exists before attempting to traverse it. Previously, if a library did not contain a proto/src directory (which is common for certain Google Cloud packages like accessapproval), the post-processor would crash with a fatal error: Uncaught UnexpectedValueException: RecursiveDirectoryIterator::__construct(././proto/src): Failed to open directory.

Fixes https://github.com/googleapis/librarian/issues/7260

### DIFF
--- a/src/PostProcessor/ProtobufNoCheckInHeaderProcessor.php
+++ b/src/PostProcessor/ProtobufNoCheckInHeaderProcessor.php
@@ -22,7 +22,11 @@ class ProtobufNoCheckInHeaderProcessor implements ProcessorInterface
 {
     public static function run(string $inputDir): void
     {
-        $protoDir = new \RecursiveDirectoryIterator($inputDir . '/proto/src');
+        $protoSrcDir = $inputDir . '/proto/src';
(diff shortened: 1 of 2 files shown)


Hectorhammett (approved):
Thanks for your contribution!

PR #857, opened 2026-08-14
fix: restore dropped code fixers and dead pagination checks

sadiqk2:
Three independent defects found while reading `src/`, each in its own commit, ordered lowest-risk first so the last one can be dropped on its own. Tests and regenerated goldens follow in their own commits at the end.

## 1. `Formatter::format()` silently discards 11–12 of its code fixers

`src/Utils/Formatter.php` appends the second half of the fixer list with the array **union** operator instead of a merge:

```php
$fixers = [ /* 11 fixers, keys 0..10 */ ];
if ($lineLength) {
    $fixers[] = self::buildLineLengthFixer($lineLength);   // key 11
}
$fixers += [ /* 25 fixers, keys 0..24 */ ];   // <-- union, not merge
```

`+` keeps the left operand's keys and adopts only right-hand keys that do not already exist. Both operands are 0-indexed lists, so the second list's leading entries collide and are dropped.

Because the optional `LineLengthFixer` changes the left operand's length, the two call paths run **different** pipelines:

| call site | fixers dropped |
|---|---|
| `format($code)` at `CodeGenerator.php:310,317,336,342`, `PhpDoc.php:425` | 11 |
| `format($code, 100)` at `CodeGenerator.php:302` (snippets) | 12 |

Dropped on both paths: `NoSpacesAfterFunctionNameFixer`, `NoEmptyCommentFixer`, `NoSpacesInsideParenthesisFixer`, `BlankLineAfterOpeningTagFixer`, `SingleImportPerStatementFixer`, `PhpdocLineSpanFixer`, `ElseifFixer`, `LineEndingFixer`, `NoTrailingWhitespaceFixer`, `NoTrailingWhitespaceInCommentFixer`, `NoBreakCommentFixer`. The snippet path additionally drops `LinebreakAfterOpeningTagFixer`.

Note that `BlankLineAfterOpeningTagFixer` is annotated in this very file as `// 1, Critical to preserving sample code.`, and it has never run. Consistent with that, none of the 82 golden snippet files under `tests/Unit/ProtoTests/*/out/samples/` has a blank line after `<?php`.

Confirmed on `php:8.2-cli` with a standalone script mirroring the exact list construction (string stand-ins for the fixer objects):

```
lineLength=NULL -> 25 fixers actually run (expected 36)  | DROPPED: 11
lineLength=120  -> 25 fixers actually run (expected 37)  | DROPPED: 12
```

Introduced by #483 (Oct 2022), the change that split the single literal in two to insert the conditional line-length fixer.

Fix: `array_merge($fixers, [...])`.

## 2. Two dead validity checks in `MethodDetails::maybeCreatePaginated()`

**`page_size` repeated-field guard is a dead store.** `$isValidPageSize = !$pageSize->isRepeated()` was immediately overwritten by both branches of the following `if/else`, so the guard never applied. A `repeated int32 page_size` passed validation and produced a client treating a `RepeatedField` as an int. The sibling `page_token` and `next_page_token` checks both include `isRepeated()`, so this reads as an omission rather than a deliberate relaxation.

**The malformed-resource-field diagnostic is unreachable, and is now deleted.** `!$resourceFieldValid` is part of the early `return null`, so the later `if (!$resourceFieldValid)` branch and its two exceptions can never be reached. Per review this takes the conservative route: the early return keeps its `!$resourceFieldValid` condition and the dead branch below is deleted. A response message whose resource field is a map, or is not first by both number and position, still generates as non-paginated, exactly as today. The dead code goes without changing what the generator accepts.

Also replaces the bitwise `&=` on a boolean with `&&` so `$resourceFieldValid` stays a bool rather than becoming int `0`/`1`.

## 3. `Map` exception messages crash on object keys

`Map` documents that keys may be any type supporting equality, and `Equality` requires only `getHash()`/`isEqualTo()`, not `__toString()`. Both error paths interpolated the key directly, so a lookup miss or duplicate-key collision on an object-keyed map raised

```
Error: Object of class X could not be converted to string
```

masking the diagnostic being reported. `ProtoCatalog::filesByService` is keyed by `ServiceDescriptorProto` and is looked up at `ResourcesGenerator.php:253`, so this is reachable. Keys are now rendered through a helper that falls back to the class name.

## Tests

Each defect now has a test that fails on `main` and passes here. 13 of them fail on `main`.

- `tests/Unit/Utils/FormatterTest.php`: one case per fixer the union operator was discarding, each asserted on both call paths, because `format($code)` and `format($code, 100)` dropped a different number of fixers. Also a case checking the line length fixer itself still runs after the merge, so the fix did not displace it.
- `tests/Unit/Generation/MethodDetailsTest.php` with `repeated-page-size.proto`: a `repeated int32 page_size` is now rejected. Before the fix the dead store let it through and the generator emitted a client treating a `RepeatedField` as an int.
- `tests/Unit/Collections/MapTest.php`: exception messages for a missing object key, a duplicate object key, a key with `__toString`, and scalar keys.

## Goldens

Regenerated, both sets.

- 162 unit golden files, via `tests/Unit/ProtoTests/GoldenUpdateMain.php`.
- 611 integration golden files, via `bazelisk run //tests/Integration:<api>_update` for all 15 APIs.

The entire diff across both sets is two things: `PhpdocLineSpanFixer` expanding one line docblocks, and `BlankLineAfterOpeningTagFixer` adding a blank line after `<?php`. No changed line in either set is anything other than a comment or a blank line, so no generated behaviour moves.

Before trusting that, I checked the regeneration is faithful. On `upstream/main` with an expunged bazel cache, regenerating all 15 integration golden sets gives a zero diff, so the toolchain I used reproduces the checked in goldens byte for byte. The same holds for the unit goldens.

One trap worth recording for whoever regenerates next. `bazel run //tests/Integration:<api>_update` will silently reuse a stale generator, because `src/` reaches the build through the `php_gapic_generator_composer_install` repository rule, and repository rules do not re-run when workspace files change. My first regeneration on `main` reproduced this branch's diff exactly, which is how I caught it. Run `bazelisk clean --expunge` first, which is what the `--expunge` flag on `run_bazel_updates.sh` is for.

## Verification

- Unit suite: 311 tests pass. Baseline on `upstream/main` is 293, and the 18 added tests account for the difference.
- Bazel integration tests: 15 of 15 pass.

An earlier revision of this description said the test suite could not be run here. That was outbound network failing to reach packagist, and it is fixed, so please disregard it. Full working setup, in case it saves anyone else the same detour:

```
composer install --prefer-dist && composer dump-autoload
git submodule update --init --recursive
USE_TOOLS_PROTOC=true vendor/bin/phpunit --bootstrap tests/Unit/autoload.php tests/Unit
```

Two things that bit me. PHP needs `ext-bcmath`, or `testGrpcServiceConfig` errors out with `Call to undefined function bccomp()`. And do not pass `--ignore-platform-reqs` to `composer install`, because it resolves Symfony 8.x, which is a parse error on PHP 8.2 and kills the run before any test reports.

## Behaviour

No commit on this branch changes what input the generator accepts.

Commit 2 is dead code removal only. The first revision revived the malformed resource field exception, which would have turned a silent fallback into a hard failure for any proto whose resource field is not first by both number and position. Following review, the unreachable branch is deleted instead. A response message whose resource field is a map, or is not first by both number and position, still generates as non paginated, exactly as today.

Commit 1's formatters do change generated output, which is why the goldens are regenerated above.

### DIFF
--- a/src/Generation/MethodDetails.php
+++ b/src/Generation/MethodDetails.php
@@ -203,9 +203,7 @@ abstract class MethodDetails
                 && $resourceByNumber[0] === $resourceByPosition[0];
         }
 
-        // A method missing any of the pagination fields is simply not paginated. A method that
-        // has them all but whose resource field is unusable is malformed, and is reported below.
-        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources)) {
+        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources) || !$resourceFieldValid) {
             return null;
         }
 
@@ -224,12 +222,6 @@ abstract class MethodDetails
         }
         if ($nextPageToken->isRepeated() || $nextPageToken->getType() !== GPBType::STRING) {
             throw new Exception('next_page_token field must be of type string.');
-        }
-        if (!$resourceFieldValid) {
-            if ($isDireGapic) {
-                throw new Exception('Item resources field must a map or repeated field.');
-            }
-            throw new Exception('Item resources field must be the first repeated field by number and position.');
         }
         return new class($svc, $desc, $outputMsg, $pageSize, $pageToken, $nextPageToken, $resources) extends MethodDetails {
             public function __construct($svc, $desc, $outputMsg, $pageSize, $pageToken, $nextPageToken, $resources)
--- a/src/Generation/MethodDetails.php
+++ b/src/Generation/MethodDetails.php
@@ -199,19 +199,22 @@ abstract class MethodDetails
 
         // Leverage short-circuting.
         if ($resourceFieldValid && !$isDireGapic) {
-            $resourceFieldValid &= !ProtoHelpers::isMap($catalog, $resources)
+            $resourceFieldValid = !ProtoHelpers::isMap($catalog, $resources)
                 && $resourceByNumber[0] === $resourceByPosition[0];
         }
 
-        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources) || !$resourceFieldValid) {
+        // A method missing any of the pagination fields is simply not paginated. A method that
+        // has them all but whose resource field is unusable is malformed, and is reported below.
+        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources)) {
             return null;
         }
 
         $isValidPageSize = !$pageSize->isRepeated();
         if ($isDireGapic) {
-            $isValidPageSize = $pageSize->getType() === GPBType::UINT32 || $pageSize->getType() === GPBType::INT32;
+            $isValidPageSize = $isValidPageSize
+                && ($pageSize->getType() === GPBType::UINT32 || $pageSize->getType() === GPBType::INT32);
         } else {
-            $isValidPageSize = $pageSize->getType() === GPBType::INT32;
+            $isValidPageSize = $isValidPageSize && $pageSize->getType() === GPBType::INT32;
         }
         if (!$isValidPageSize) {
             throw new Exception('page_size field must be of type ' . ($isDireGapic ? 'uint32 or int32' : 'int32') . '.');
--- a/src/Collections/Map.php
+++ b/src/Collections/Map.php
@@ -63,7 +63,7 @@ class Map implements \IteratorAggregate, \Countable, \ArrayAccess
         $data = [];
         foreach ($pairs as [$k, $v]) {
             if (static::apply($data, $k, 1, $v)[0]) {
-                throw new Exception("Cannot add two items with the same key $k");
+                throw new Exception('Cannot add two items with the same key ' . static::keyToString($k));
             }
         }
         return new Map($data, count($pairs));
@@ -155,7 +155,25 @@ class Map implements \IteratorAggregate, \Countable, \ArrayAccess
         }
         $dbt = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
         $caller = isset($dbt[1]['function']) ? $dbt[1]['function'] : null;
-        throw new Exception("Key $key does not exist, called from $caller");
+        throw new Exception('Key ' . static::keyToString($key) . " does not exist, called from $caller");
+    }
+
+    /**
+     * Render a key for use in an exception message. Keys may be any type that supports
+     * equality, which does not imply they are convertible to string; interpolating such a
+     * key directly would raise an "object could not be converted to string" error, masking
+     * the diagnostic being reported.
+     *
+     * @param mixed $key The key to render.
+     *
+     * @return string
+     */
+    private static function keyToString($key): string
+    {
+        if (is_object($key)) {
+            return method_exists($key, '__toString') ? (string) $key : '<' . get_class($key) . '>';
+        }
+        return (string) $key;
     }
 
     /** @inheritDoc */
--- a/src/Utils/Formatter.php
+++ b/src/Utils/Formatter.php
@@ -72,7 +72,7 @@ class Formatter
             $fixers[] = self::buildLineLengthFixer($lineLength);
         }
 
-        $fixers += [
+        $fixers = array_merge($fixers, [
             new Fixer\FunctionNotation\NoSpacesAfterFunctionNameFixer(), // 2, PSR2
             new Fixer\Comment\NoEmptyCommentFixer(), // 2
             new Fixer\Whitespace\NoSpacesInsideParenthesisFixer(), // 2, PSR2
@@ -98,7 +98,7 @@ class Formatter
             new Fixer\Whitespace\ArrayIndentationFixer(), // -31
             new Fixer\Whitespace\MethodChainingIndentationFixer(), // -34
             new Fixer\Whitespace\SingleBlankLineAtEofFixer(), // -50, PSR2 (must run last)
-        ];
+        ]);
 
         // Fixer temporarily removed:
         // new Fixer\Whitespace\BlankLineBeforeStatementFixer(), // -21
--- a/tests/Integration/goldens/asset/samples/V1/AssetServiceClient/analyze_iam_policy.php
+++ b/tests/Integration/goldens/asset/samples/V1/AssetServiceClient/analyze_iam_policy.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * Copyright 2026 Google LLC
  *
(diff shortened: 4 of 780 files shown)


src/Generation/MethodDetails.php
```diff
@@ -168,19 +168,22 @@ private static function maybeCreatePaginated(ServiceDetails $svc, MethodDescript
...
+            $resourceFieldValid = !ProtoHelpers::isMap($catalog, $resources)
                 && $resourceByNumber[0] === $resourceByPosition[0];
         }
 
-        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources) || !$resourceFieldValid) {
+        // A method missing any of the pagination fields is simply not paginated. A method that
+        // has them all but whose resource field is unusable is malformed, and is reported below.
+        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources)) {
             return null;
         }
```
cy-yun: As you noted in the PR description, making this exception reachable turns a silent fallback into a hard failure. To avoid breaking existing clients that might currently rely on this fallback behavior, let's restore the `|| !$resourceFieldValid` here.

```suggestion
        // A method missing any of the pagination fields is simply not paginated. A method that
        // has them all but whose resource field is unusable is malformed, and is reported below.
        if (is_null($pageSize) || is_null($pageToken) || is_null($nextPageToken) || is_null($resources) || !$resourceFieldValid) {
            return null;
        }

File: src/Generation/MethodDetails.php
Comment (cy-yun) [inline src/Generation/MethodDetails.php]: Since we are keeping the early return above, this block remains unreachable. We can safely delete it entirely to remove the dead code without changing what the generator accepts.

```suggestion
```

cy-yun (changes requested):
Great catch on these issues! The fixes for `Formatter.php` and `Map.php` look solid, as do the dead store and bitwise boolean fixes in `MethodDetails.php`. 

Regarding the pagination check, I've left a couple of inline comments to go with the conservative route you suggested (deleting the unreachable branch instead of reviving it). This ensures we don't accidentally break generation for any existing protos that were silently falling back to being unpaginated. Once those tweaks are made, this looks good to go!

src/Generation/MethodDetails.php
sadiqk2: Restored in `63e94fda`, along with removing the comment I had added above it. With the condition back in the early return, that comment was describing the wrong behaviour.

src/Generation/MethodDetails.php
sadiqk2: Removed in `63e94fda`. With the early return unchanged, the branch is unreachable, so deleting it gives the dead code fix without any behaviour change.

sadiqk2:
Thanks. I have taken the conservative route on both points in `63e94fda`. The early return keeps `|| !$resourceFieldValid`, the unreachable branch below is removed, and I have dropped the comment I had written above the return, since it was describing behaviour that is no longer there. The dead store fix on `$isValidPageSize` and the `&=` to `&&` change are untouched.

Net effect of that commit: `MethodDetails` accepts exactly the same protos it accepts on main, except the `repeated page_size` case, which the dead store was letting through into a client that treats a `RepeatedField` as an int.

Two points from the description still hold, as neither is affected by this change:

1. The test suite has still not been run here. Composer dependencies do not install in my environment, so the only check on the changed files is `php -l`. The rest is from reading the code.
2. Golden files are not regenerated. The `Formatter` commit turns on 11 to 12 fixers that have not run since #483, so generated output will reformat and presubmit is expected to fail on this branch until the goldens are updated. I am happy to push a regeneration, or to split that commit into a separate PR so the other two can go in clean. Kindly let me know which you prefer.

Hectorhammett (changes requested):
I read the comment about you not being able to run the unit tests and I am curious about what is the reason for it?

I went ahead and ran the unit tests in my environment and they are failing, so this seems like it has work to do.

Currently I can see what are we trying to accomplish and I regenerated the Golden clients and it does look good, but the failure to include any other unit tests to test the bugs you are trying to accomplish makes this a bit hard to merge.

Why aren't you able to install the dependencies? Is there any error coming up? 

I would suggest to add unit tests to test what you are trying to fix.

sadiqk2:
@Hectorhammett thanks for actually running it, and sorry you had to.

On why the dependencies would not install: outbound network from my environment could not reach packagist at the time I opened this, so `composer install` timed out on `repo.packagist.org/packages.json`. Nothing to do with the repo. That is working again, so I have run everything now.

For the record, in case it saves someone else the same detour, the setup that works is:

```
composer install --prefer-dist && composer dump-autoload
git submodule update --init --recursive
USE_TOOLS_PROTOC=true vendor/bin/phpunit --bootstrap tests/Unit/autoload.php tests/Unit
```

Two things bit me after that. PHP needs `ext-bcmath` or `testGrpcServiceConfig` errors with `Call to undefined function bccomp()`. And do not pass `--ignore-platform-reqs` to `composer install`, because it resolves Symfony 8.x, which is a parse error on PHP 8.2 and kills the run before any test reports. That one looks like a broken repo when it is not.

On the tests, you were right that there were none. Each fix now has one that fails on `main` and passes here, 13 in total:

- `tests/Unit/Utils/FormatterTest.php`, one case per fixer the union operator was discarding. Each case runs on both call paths, because `format($code)` and `format($code, 100)` dropped a different number of fixers. There is also a case checking the line length fixer itself still runs after the merge, so the fix does not displace it.
- `tests/Unit/Generation/MethodDetailsTest.php` with a `repeated-page-size.proto` fixture. A `repeated int32 page_size` is rejected now. Before the fix the dead store let it through and the generator emitted a client treating a `RepeatedField` as an int.
- `tests/Unit/Collections/MapTest.php`, exception messages for a missing object key, a duplicate object key, a key with `__toString`, and scalar keys.

On the goldens, you said yours looked good and mine agree. Both sets are regenerated on the branch now: 162 unit files and 611 integration files across all 15 APIs. The whole diff is two things, `PhpdocLineSpanFixer` expanding one line docblocks and `BlankLineAfterOpeningTagFixer` adding a blank line after `<?php`. I checked every changed line and none of them is anything other than a comment or a blank, so no generated behaviour moves.

Before trusting that I checked the regeneration is faithful. On `upstream/main` with an expunged bazel cache, regenerating all 15 integration golden sets gives a zero diff, so my toolchain reproduces the checked in goldens byte for byte.

One trap worth writing down for next time. `bazel run //tests/Integration:<api>_update` will silently reuse a stale generator, because `src/` reaches the build through the `php_gapic_generator_composer_install` repository rule, and repository rules do not re-run when workspace files change. My first regeneration on `main` reproduced this branch's diff exactly, which is how I noticed. `bazelisk clean --expunge` first, which is what the `--expunge` flag on `run_bazel_updates.sh` is for.

Current state: unit suite 311 passing against 293 on `upstream/main`, and 15 of 15 bazel integration tests passing. I have also rewritten the description so it reports what was measured instead of what I expected.

